PR #12, opened 2018-05-24
Misc fixes

nyalldawson:
Fix some warnings and clang analyzer issues


nyalldawson:
@PeterPetrik any idea why the astyle test fails here?

PeterPetrik:
@nyalldawson thanks a lot for fixes. With astyle, I added extra output on the other PR, so let me investigate/fix that. My problem was that it uses astyle 3.0.1

PR #40, opened 2018-10-22
frmts/ascii_dat: Replace tabs by spaces

michelk:
Basement v2.8 writes results with tabs e.g.

    NAME "VertexScalarDataset"$
    TIMEUNITS se$
    TS 0»···0.000000$
    1·»·$
    2·»·$
    3·»·$
    2·»·$
    1·»·$

This wasn't parsed correctly by the ascii_dat parser. Therefor tabs are
now replaced by spaces when parsing the simulation results.

### DIFF
--- a/mdal/frmts/mdal_ascii_dat.cpp
+++ b/mdal/frmts/mdal_ascii_dat.cpp
@@ -87,6 +87,10 @@ void MDAL::LoaderAsciiDat::load( MDAL::Mesh *mesh, MDAL_Status *status )
 
   while ( std::getline( in, line ) )
   {
+    // Replace tabs by spaces,
+    // since basement v.2.8 uses tabs instead of spaces (e.g. 'TS 0\t0.0')
+    line = replace( line, "\t", " " );
+
     std::vector<std::string> items = split( line,  " ", SplitBehaviour::SkipEmptyParts );
     if ( items.size() < 1 )
       continue; // empty line?? let's skip it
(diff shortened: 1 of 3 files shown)


mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -87,6 +87,11 @@ void MDAL::LoaderAsciiDat::load( MDAL::Mesh *mesh, MDAL_Status *status )
 
   while ( std::getline( in, line ) )
   {
+    //Replace tabs by spaces
+    for(std::string::size_type i = 0; i < line.size(); ++i) {
```
PeterPetrik: please add a small comment that basement 2.8 uses tabs instead of spaces

Maybe use `line = mdal::replace(line, 't', '  ')`?
https://github.com/lutraconsulting/MDAL/blob/master/mdal/mdal_utils.hpp#L45

PeterPetrik:
@michelk thank you for your contribution! The PR looks very good. GitHub is broken today, so CI tests do not run. I think the PR fails the code style check since there are missing some whitespaces. Can you please run `scrips/check_all.sh`. Best to add code check to pre-commit git hooks. 

If you have some small (<1MB) example file of basement 2.8, you can add it to the example suite instead of the example you added, but it is not required.

michelk:
Hi Peter, 
I just adapted the code for your suggestions.

Thanks, Michel

Issue #13, opened 2018-05-28
Lazy loading of XMDF dataset


PeterPetrik:
specially xmdf (hdf5) format datasets should be loaded only when requested.

PR #45, opened 2018-11-22
[API Break] XMDF lazy

PeterPetrik:
fixes https://github.com/lutraconsulting/MDAL/issues/13

### DIFF
--- a/mdal/frmts/mdal_xmdf.hpp
+++ b/mdal/frmts/mdal_xmdf.hpp
@@ -20,6 +20,48 @@
 namespace MDAL
 {
 
+  /**
+   * The XmdfDataset reads the data directly from HDF5 file
+   * by usage of hyperslabs retrieval
+   *
+   * basically all (timesteps) data for one particular dataset groups
+   * are stored in single 2D array with structure
+   *
+   * timestep1 x1 y1 x2 y2 ...... xN yN
+   * timestep2 x1 y1 x2 y2 ...... xN yN
+   * ...
+   * timestepN x1 y1 x2 y2 ...... xN yN
+   *
+   * or for scalar datasets
+   *
+   * timestep1 x1 x2 ...... xN
+   * timestep2 x1 x2 ...... xN
+   * ...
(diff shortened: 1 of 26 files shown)


mdal/frmts/mdal_xmdf.hpp
```diff
@@ -20,13 +20,56 @@
...
+      size_t vectorData( size_t indexStart, size_t count, double *buffer ) override;
+      size_t activeData( size_t indexStart, size_t count, char *buffer ) override;
+
+      const HdfDataset &dsValues() const;
+      const HdfDataset &dsActive() const;
+      hsize_t timeIndex() const;
+
+    private:
+      HdfDataset mHdf5DatasetValues;
+      HdfDataset mHdf5DatasetActive;
```
wonder-sk: are these some simple pointers? I'm wondering if keeping them open for each dataset is a good practice, i.e. if they do not force hdf library to consume some extra resources (even though we may not need to used those datasets)

mdal/mdal_data_model.cpp
```diff
@@ -8,20 +8,91 @@
...
+  for ( size_t i = 0; i < count; ++i )
+  {
+    const MDAL::Value value = values[ indexStart + i ];
+    if ( value.noData )
+    {
+      buffer[i] = NODATA;
+    }
+    else
+    {
+      buffer[i] = value.x;
```
wonder-sk: Maybe for scalar datasets we do not need to keep x+y to save some memory?

mdal/frmts/mdal_xmdf.hpp
PeterPetrik: they are shared pointers, and there is one pointer for ALL timesteps in one group, so basically we should not run to exotic numbers here

mdal/mdal_data_model.cpp
PeterPetrik: it should be reworked a lot, but I would keep it for separate PR/task since it is not related strictly to the lazy loading

PR #46, opened 2018-11-27
Mesh stream + XMDF lazy

PeterPetrik:


mdal/api/mdal.h
```diff
@@ -94,41 +96,54 @@ MDAL_EXPORT int MDAL_M_vertexCount( MeshH mesh );
...
 //! Returns dataset group handle
 MDAL_EXPORT DatasetGroupH MDAL_M_datasetGroup( MeshH mesh, int index );
 
+///////////////////////////////////////////////////////////////////////////////////////
+/// MESH VERTICES
+///////////////////////////////////////////////////////////////////////////////////////
+
+//! Returns iterator to the mesh vertices
+//! For some formats this may effectively load all vertices in-memory unit iterator is closed
+MDAL_EXPORT MeshVerticesIteratorH MDAL_M_readVertices( MeshH mesh );
```
wonder-sk: MDAL_M_vertexIterator() ?
MeshVertexIteratorH ?

mdal/api/mdal.h
```diff
@@ -94,41 +96,54 @@ MDAL_EXPORT int MDAL_M_vertexCount( MeshH mesh );
...
+//! Returns vertices from iterator for the mesh
+//! \param iterator mesh data iterator
+//! \param verticesCount maximum number or vertices to be written to buffer
+//! \param coordinates must be allocated to 3* verticesCount items to store x1, y1, z1, ..., xN, yN, zN coordinates
+//! \returns number of vertices written in the buffer
+MDAL_EXPORT int MDAL_M_vertices( MeshVerticesIteratorH iterator, int verticesCount, double *coordinates );
```
wonder-sk: MDAL_VI_next() ?

mdal/api/mdal.h
```diff
@@ -94,41 +96,54 @@ MDAL_EXPORT int MDAL_M_vertexCount( MeshH mesh );
...
+//! Closes mesh data iterator, frees the memory
+MDAL_EXPORT void MDAL_M_CloseMeshVerticesIterator( MeshVerticesIteratorH iterator );
```
wonder-sk: MDAL_VI_close() ? (destroy?)

mdal/api/mdal.h
```diff
@@ -94,41 +96,54 @@ MDAL_EXPORT int MDAL_M_vertexCount( MeshH mesh );
...
+
+///////////////////////////////////////////////////////////////////////////////////////
+/// MESH FACES
+///////////////////////////////////////////////////////////////////////////////////////
+
+//! Returns iterator to the mesh vertices
```
wonder-sk: s/vertices/faces/

mdal/api/mdal.h
```diff
@@ -94,41 +96,54 @@ MDAL_EXPORT int MDAL_M_vertexCount( MeshH mesh );
...
+//! For some formats this may effectively load all vertices in-memory unit iterator is closed
```
wonder-sk: s/vertices/faces/
s/unit/until/

mdal/api/mdal.h
```diff
@@ -94,41 +96,54 @@ MDAL_EXPORT int MDAL_M_vertexCount( MeshH mesh );
...
+MDAL_EXPORT MeshFacesIteratorH MDAL_M_readFaces( MeshH mesh );
```
wonder-sk: renaming like for vertex iterator?
MeshFaceIteratorH?

mdal/api/mdal.h
```diff
@@ -94,41 +96,54 @@ MDAL_EXPORT int MDAL_M_vertexCount( MeshH mesh );
...
+//! Returns next faces from iterator for the mesh
+//! \param iterator mesh data iterator
+//! \param facesCount maximum number or faces to be written to buffer
+//! \param faceVerticesStartIndex array must be allocated to be able to store facesCount + 1 items
```
wonder-sk: why +1 items? (we don't need need to write zero for the first item, and it would avoid "off by one" errors)

mdal/api/mdal.h
```diff
@@ -94,41 +96,54 @@ MDAL_EXPORT int MDAL_M_vertexCount( MeshH mesh );
...
+//! \param facesVertexIndicesCount number of vertex indexes buffer can store, minimum is MDAL_M_faceVerticesMaximumCount()
+//! \param facesVertexIndices writes vertex indexes for faces
+//! \returns number of faces written in the buffer
+MDAL_EXPORT int MDAL_M_faces( MeshFacesIteratorH iterator,
+                              int facesCount,
+                              int *faceVerticesStartIndex,
+                              int facesVertexIndicesCount,
+                              int *facesVertexIndices );
```
wonder-sk: I have trouble connecting that `facesCount` and `faceVerticesStartIndex` arguments work together as they have completely different names. And I need to read everything twice to remember which of `faceVerticesStartIndex` and `facesVertexIndices` means what. How about e.g.:
- vertexIndicesBuffer
- vertexIndicesBufferLen
- faceOffsetsBuffer
- faceOffsetsBufferLen

wonder-sk: It is worth noting in API doc that reading stops when vertexIndicesBuffer capacity is full / faceOffsetsBuffer capacity is full / end of faces is reached, whatever comes first.

mdal/api/mdal.h
```diff
@@ -156,25 +205,36 @@ MDAL_EXPORT double MDAL_D_time( DatasetH dataset );
...
+//!
+//! \param dataset handle to dataset
+//! \param indexStart index of face/vertex to start reading of values to the buffer
+//! \param count number of values to be written to the buffer
+//! \param dataType type of values to be written to the buffer
+//! \param buffer output array to be populated with the values. must be already allocated
+//!               For SCALAR_DOUBLE, the minimum size must be valuesCount * size_of(double)
+//!               For VECTOR_2D_DOUBLE, the minimum size must be valuesCount * 2 * size_of(double).
+//!                                     Values are returned as x1, y1, x2, y2, ..., xN, yN
+//!               For ACTIVE_BOOL, the minimum size must be valuesCount * size_of(char)
```
wonder-sk: actually looking at the implementation it is sizeof(int) not sizeof(char) ...?

mdal/api/mdal.h
```diff
@@ -156,25 +205,36 @@ MDAL_EXPORT double MDAL_D_time( DatasetH dataset );
...
+//! \returns number of values written to buffer. If return value != count requested, see MDAL_LastStatus() for error type
+MDAL_EXPORT int MDAL_D_data( DatasetH dataset, int indexStart, int count, MDAL_DataType dataType, void *buffer );
+
+//! Returns the min and max values of the dataset
+//! Returns NaN on error
+MDAL_EXPORT void MDAL_D_minimumMaximum( DatasetH dataset, double *min, double *max );
```
wonder-sk: don't we need just min/max for a whole dataset group?

mdal/mdal_data_model.hpp
```diff
@@ -27,45 +29,40 @@ namespace MDAL
...
     double y;
-    double z; // Bed elevation
-  } Vertex;
-
-  typedef std::vector<size_t> Face;
 
-  typedef std::vector<Vertex> Vertices;
-  typedef std::vector<Face> Faces;
+    bool noData = false;
+  } Value; //Dataset Value
```
wonder-sk: is this Value struct still used anywhere?

mdal/mdal_data_model.hpp
```diff
@@ -74,42 +71,94 @@ namespace MDAL
...
       std::string name();
       void setName( const std::string &name );
 
       Metadata metadata;
-
-      bool isScalar = true;
-      bool isOnVertices = true;
       Datasets datasets;
-      std::string uri; // file/uri from where it came
+      Mesh *parent = nullptr;
```
wonder-sk: move to private as well and have getter/setter?

mdal/mdal_data_model.hpp
```diff
@@ -74,42 +71,94 @@ namespace MDAL
...
+            const std::string &uri );
+      virtual ~Mesh();
+
+      void setSourceCrs( const std::string &str );
+      void setSourceCrsFromWKT( const std::string &wkt );
+      void setSourceCrsFromEPSG( int code );
+
+      void setVerticesCount( const size_t &verticesCount );
+      void setFacesCount( const size_t &facesCount );
+      void setFaceVerticesMaximumCount( const size_t &faceVerticesMaximumCount );
```
wonder-sk: better to pass simple values by value than by const ref

mdal/mdal_memory_data_model.cpp
```diff
@@ -0,0 +1,209 @@
...
+
+    const MDAL::Value value = values[ indexStart + i ];
+    if ( value.noData )
+    {
+      buffer[i] = MDAL_NAN;
+    }
+    else
+    {
+      buffer[i] = value.x;
+    }
```
wonder-sk: why do we need this `if (nodata)` test here? can't we just have "values" array filled up with NaN values where needed?

mdal/mdal_memory_data_model.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+      size_t scalarData( size_t indexStart, size_t count, double *buffer ) override;
+      size_t vectorData( size_t indexStart, size_t count, double *buffer ) override;
+      size_t activeData( size_t indexStart, size_t count, int *buffer ) override;
+
+      /**
+       * size - face count if !isOnVertices
+       * size - vertex count if isOnVertices
+       */
+      std::vector<Value> values;
```
wonder-sk: I would suggest to have two different arrays - one for scalars and one for vectors. Like this we are wasting a lot of memory.

Also, noData in Value seems unneeded and bloats the structure (8 + 8 + 4 bytes + 4 bytes alignment (I think)). How about:
```
std::vector<double> valuesScalar;
struct double2d { double x, double y };
std::vector<double2d> valuesVector;
```
This would even allow us to use simple memcpy to fill up the buffers.

mdal/frmts/mdal_2dm.cpp
```diff
@@ -17,6 +17,46 @@
...
+}
+
+size_t MDAL::Mesh2dm::vertexIndex( size_t vertexID ) const
+{
+  auto ni2i = mVertexIDtoIndex.find( vertexID );
+  if ( ni2i != mVertexIDtoIndex.end() )
+  {
+    return  ni2i->second; // convert from ID to index
+  }
+  return vertexID;
```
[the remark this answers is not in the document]
PeterPetrik: it is a bit reworked, `mVertexIDtoIndex` now stores ID's that are mapped to different index than `id-1`. So for most of the files the `mVertexIDtoIndex` is just empty map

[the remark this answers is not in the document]
wonder-sk: ah ok - a comment about that would be nice :)

mdal/mdal_data_model.hpp
PeterPetrik: that is how it ends up when you trust automatic code creation tools ...

mdal/mdal_memory_data_model.hpp
PeterPetrik: @wonder-sk or something like this?
```
  class Values {
    public:
      Values(size_t nItems, bool isVector);

      void setX(size_t index, double val);
      void setY(size_t index, double val); // only for isVector = true

      double x(size_t index) const;
      double y(size_t index) const; // only for isVector = true

      double* data();
      double* constData() const;
    private:
      bool mIsVector;
      // nItems * 2 for vectors
      // nItems for scalars
      // x=NaN, y=NaN for NODATA values
      std::vector<double> mValues;
  };
```

wonder-sk: Hmm... I don't like the idea to have a class with X/Y getters/setters when that only makes sense for vectors. I would rather go with just simple `std::vector<double> value` then. Or to have two separate classes - e.g. ScalarValues / Vector2DValues if we want full classes (but we probably don't need that).

PeterPetrik:
@wonder-sk all should be fixed and prepared for next round of reviews

mdal/api/mdal.h
PeterPetrik: In qgis code there is some place when there is query for dataset min/max...

mdal/frmts/mdal_3di.cpp
```diff
@@ -93,21 +93,21 @@ void MDAL::Loader3Di::populateFacesAndVertices( MDAL::Mesh *mesh )
...
   // are used in multiple faces
-  mDimensions.setDimension( CFDimensions::Vertex2D, mesh->vertices.size() );
+  mDimensions.setDimension( CFDimensions::Vertex2D, vertices.size() );
 }
 
 void MDAL::Loader3Di::addBedElevation( MDAL::Mesh *mesh )
 {
   assert( mesh );
-  if ( mesh->faces.empty() )
+  if ( 0 == mesh->facesCount() )
```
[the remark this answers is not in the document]
PeterPetrik: faces vector is private and only for memory group/dataset.. facesCount is for any type of dateset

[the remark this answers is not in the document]
wonder-sk: I just meant why `0 == x` rather than `x == 0` :)

PR #71, opened 2019-01-08
Load all groups from XMDF file

PeterPetrik:
try to load all groups, not just hardcoded "Temporal, Maximums and Difference"


mdal/frmts/mdal_xmdf.hpp
```diff
@@ -58,6 +58,26 @@ namespace MDAL
...
+       *     - ..
+       *   - Maximums
+       *     - Depth
+       *     - Velocity
+       *     - ..
+       *   - Difference (res_to_res.exe TUFLOW utility tool)
+       *     - ..
+       *   - Times (e.g. time of peak velocity)
+       *     - ..
+       *   - ...
```
wonder-sk: Maybe worth saying that this is the structure used by TUFLOW but other solvers may use a different structure within XMDF?

PR #105, opened 2019-03-20
fix #104 GDAL datasets without NODATA set

### LINKED ISSUE #104: GDAL based data without NODATA value
if gdal-based dataset is without NODATA value set, it is automatically set to 0 in MDAL and some values could be misinterpreted.

related to https://issues.qgis.org/issues/21537

PeterPetrik:
fix #104


mdal/frmts/mdal_gdal.cpp
```diff
@@ -349,7 +355,8 @@ void MDAL::DriverGdal::addDataToOutput( GDALRasterBandH raster_band, std::shared
     {
       unsigned int idx = x + mXSize * y;
       double val = mPafScanline[x];
-      if ( !MDAL::equals( val, nodata ) )
+
+      if ( std::isnan( nodata ) || !MDAL::equals( val, nodata ) )
```
wonder-sk: It took me a bit to understand what is going on... maybe it would be nicer to have bool "hasNoData" variable instead of check for isnan(nodata)

PeterPetrik: +1

Issue #31, opened 2018-09-27
Support telemac outputs


saberraz:

PeterPetrik:
- [Serafin files](http://www.gdal.org/drv_selafin.html): Outputs of the TELEMAC 2D hydrodynamic modelling package

saberraz:
Maybe @patricev can help?

ccoulet:
Hi guys
I started to work on the selafin (or serafin) driver to support Telemac in mdal.
I think I could know read a mesh but unfortunately, I was not able to test it in QGIS...
I compile on windows with mingw and obtain a file libmdal.dll which is, I suppose, my version of mdalprovider.dll
But when replacing the original version by mine in QGIS, I have a message saying that the module loading failed (specified module not found)
Am I missing something?
Regards

saberraz:
@ccoulet that sounds very promising!

You need also to build QGIS without internal mdal (-D WITH_INTERNAL_MDAL=FALSE).

If you do a PR on what you have, I can give it a try....

ccoulet:
I didn't planned to build QGIS...
And I think it will take time to build it on my Windows computer...
I will probably try first on a Linux one, this will be probably easier...
In the mean time I forked MDAL on https://github.com/Artelia/MDAL and create a branch Telemac in order to let you test it...
Maybe it will be simpler if you create a Telemac branch on the MDAL repository in which I could commit my update...

saberraz:
Managed to compile it successful (telemac branch) but get error when trying to open this file:
[baxter_SS_hotstart.slf.zip](https://github.com/lutraconsulting/MDAL/files/2920340/baxter_SS_hotstart.slf.zip)

(worth adding a small test file too).

ccoulet:
Thanks for the feed back. Any error message?
I added a slf directory in test with a mesh file in it
I have a problem in the compilation of gmock which fail so cannot work on a test

PeterPetrik:
Hi @ccoulet , when you will have some working code (ideally with a small test file in tests/data/ and test_telemac.cpp), please do pull request and I will review the code. Thanks for contribution!

As Saber mentioned, you need to compile QGIS yourself without internal MDAL to be able to test it in QGIS.

ccoulet:
Hi @PeterPetrik and @saberraz 
Is there a solution for you to share with me a QGIS version without internal MDAL? I don't have too much time and compiling QGIS seems a little be complex as I never done it before.
[...]

PR #120, opened 2019-05-01
Support Selafin/Serafin/Telemac format

PeterPetrik:
fix https://github.com/lutraconsulting/MDAL/issues/31


saberraz:
@ccoulet would it be possible to provide a small dataset for test. It should have mesh and vector data ideally. Thanks

PeterPetrik:
> @ccoulet would it be possible to provide a small dataset for test. It should have mesh and vector data ideally. Thanks

I have used the test from other PR for scalar dataset. We are missing vector data (velocity u and velocity v arrays) and dataset written with double precision "SERAFIND" instead or "SERAFIN". Nevetheless, we have some basic test coverage even without those.

PR #124, opened 2019-05-22
fix #123 maximum datasets for flo-2d

### LINKED ISSUE #123: [flo-2d] Maximum datasets are wrongly defined on vertices
so the maximums do not match with the real data

PeterPetrik:


PeterPetrik:
@saberraz can you please test?

saberraz:
@PeterPetrik it all looks good. I tested your branch and the max values appear in the correct place.

PR #127, opened 2019-06-19
Update mdal_selafin.cpp

ccoulet:
Small addition to add the capabilities to detect vector variable even the variable name is in french.


PeterPetrik:
can you please add a test to the test suite with such file? Thanks

ccoulet:
I will do it.
Should I create a test_selafin2.cpp file or could I add a second test bloc in test_selafin.cpp file.
I plan to use the Malpasset result file (in french) used for reference

PeterPetrik:
best to add to existing one .

if you can fix the "old" format also it would be good.

PR #128, opened 2019-06-19
Update of the previous pull request with the test

ccoulet:
I closed the previous pull request and made this new one to complete with the addition of a test
A new selafin result file was added
Some tests are also added but should probably be checked as I'm not sure of the time and variable management in the dataset...


PeterPetrik:
please fix the CI, looks like there is a compilation error

PeterPetrik:
@ccoulet thanks for this PR. Is it common to have the variable names in French for this format?

ccoulet:
As Telemac is a french software, all results could have the variable names in French or in English. It's a user choice...

Issue #121, opened 2019-05-03
Editing/Creating Mesh 


PeterPetrik:
Right now we have only possiblity to load mesh, but we are not able to edit it. Similarly we have API for dataset to both read and write. 

It would be nice to add API to be able to create empty mesh, add some vertices and faces to it.  For beginning we can implement 2dm files writer, so we would be able to persist the new/edited mesh back to 2dm file.

vcloarec:
Have you chosen the 2dm format because it is the simplest ?

PeterPetrik:
yep, 2dm is the ASCII format, well defined and easiest for beginning.

vcloarec:
Ok, but is this file format is the one which have the most interest to be edited ? 
I don't know this format, but is this format is needed to be created and edited by user ?
Just questions, I really don't know.

vcloarec:
What ate the TIN file formats used by hydraulic models ?

PeterPetrik:
The API for write will be the same for all drivers that have capability to write/edit meshes. To begin actual implementation, I think it is easiest to start with 2dm since it is ASCII format and one can check result in text editor

2dm is used in various models , e.g. TUFLOW, HECRAS, BASEMENT, ...

vcloarec:
Back to editing mesh after couple weeks, I have implemented a simple 2dm writer. #

vcloarec:
#

vcloarec:
Editing mesh (e.g. add vertices or faces) needs to have a good handle of the vertices and faces arrangement and it is not trivial. I don't know if the aim of MDAL is to handle this arrangements... If yes, there is a lot of work to do.
If not necessary, there is a need for third library to handle properly this arrangement, but license is may be not compatible to make MDAL dependent from it.
I am thinking about [CGAL](https://doc.cgal.org/latest/Manual/packages.html#PartTriangulationsAndDelaunayTriangulations) or [Jonathan Richard Shewchuk's code](https://www.cs.cmu.edu/~quake/triangle.html) to generate triangulation.
An other way could may be that MDAL offers sort of interface to derive from some generator/arrangements engines. So the users can implement engines to generate/edit mesh and use MDAL to open/save and to display mesh with QGIS.
What do you think about?

PeterPetrik:
I think that this ticket assumes that you create vertices/faces externally and MDAL just have api to modify the mesh frame. So in first stage I assume we do not need triangulation library.

PR #133, opened 2019-07-01
2dm files writer

vcloarec:
add a simple 2dm files writer. issue #121

### DIFF
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -56,6 +56,7 @@ enum MDAL_Status
   Err_IncompatibleDatasetGroup,
   Err_MissingDriver,
   Err_MissingDriverCapability,
+  Err_FailToWriteToDisk,
   // Warnings
   Warn_UnsupportedElement,
   Warn_InvalidElements,
@@ -120,6 +121,10 @@ MDAL_EXPORT const char *MDAL_DR_filters( DriverH driver );
 //! This may effectively load whole mesh in-memory for some providers
 //! Caller must free memory with MDAL_CloseMesh() afterwards
 MDAL_EXPORT MeshH MDAL_LoadMesh( const char *meshFile );
+
+MDAL_EXPORT void MDAL_SaveMesh( MeshH mesh, const char *meshFile );
+
+
 //! Closes mesh, frees the memory
 MDAL_EXPORT void MDAL_CloseMesh( MeshH mesh );
 //! Returns mesh projection
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -100,6 +100,9 @@ MDAL_EXPORT bool MDAL_DR_meshLoadCapability( DriverH driver );
 //! Returns whether driver has capability to write/edit dataset (groups)
 MDAL_EXPORT bool MDAL_DR_writeDatasetsCapability( DriverH driver );
 
+//! Returns whether driver has capability to save mesh
+MDAL_EXPORT bool MDAL_DR_SaveMeshCapability( DriverH driver );
+
 //! Returns name of MDAL driver
 //! not thread-safe and valid only till next call
 MDAL_EXPORT const char *MDAL_DR_name( DriverH driver );
@@ -122,11 +125,13 @@ MDAL_EXPORT const char *MDAL_DR_filters( DriverH driver );
 //! Caller must free memory with MDAL_CloseMesh() afterwards
 MDAL_EXPORT MeshH MDAL_LoadMesh( const char *meshFile );
 
-MDAL_EXPORT void MDAL_SaveMesh( MeshH mesh, const char *meshFile );
-
-
 //! Closes mesh, frees the memory
 MDAL_EXPORT void MDAL_CloseMesh( MeshH mesh );
+
+//! Save mesh on a file whith the specified driver. On error see MDAL_LastStatus for error type.
+//! Save the datasets if the driver has the capability
+MDAL_EXPORT void MDAL_SaveMesh( MeshH mesh, const char *meshFile, const char *driver );
+
 //! Returns mesh projection
 //! not thread-safe and valid only till next call
 MDAL_EXPORT const char *MDAL_M_projection( MeshH mesh );
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -128,8 +128,9 @@ MDAL_EXPORT MeshH MDAL_LoadMesh( const char *meshFile );
 //! Closes mesh, frees the memory
 MDAL_EXPORT void MDAL_CloseMesh( MeshH mesh );
 
-//! Save mesh on a file whith the specified driver. On error see MDAL_LastStatus for error type.
-//! Save the datasets if the driver has the capability
+//! Saves mesh on a file with the specified driver. On error see MDAL_LastStatus for error type.
+//! Saves the datasets if the driver has the capability (as, right now, only the 2DM format is implemented for saving,
+//! it does not save the datasets except the terrain elevation of the mesh (Z value of vertices)).
 MDAL_EXPORT void MDAL_SaveMesh( MeshH mesh, const char *meshFile, const char *driver );
 
 //! Returns mesh projection
--- a/mdal/frmts/mdal_2dm.cpp
+++ b/mdal/frmts/mdal_2dm.cpp
@@ -309,11 +309,14 @@ void MDAL::Driver2dm::save( const std::string &uri, MDAL::Mesh *mesh, MDAL_Statu
     vertexIterator->next( 1, vertex );
     line = "ND ";
     line.append( std::to_string( i + 1 ) );
-    for ( size_t j = 0; j < 3; ++j )
+    for ( size_t j = 0; j < 2; ++j )
     {
       line.append( " " );
-      line.append( std::to_string( vertex[j] ) );
+      line.append( MDAL::coordinateToString( vertex[j] ) );
     }
+    line.append( " " );
+    line.append( MDAL::doubleToString( vertex[2] ) );
+
     file << line << std::endl;
   }
 
--- a/mdal/frmts/mdal_2dm.cpp
+++ b/mdal/frmts/mdal_2dm.cpp
@@ -82,7 +82,7 @@ MDAL::Driver2dm::Driver2dm():
   Driver( DRIVER_NAME,
           "2DM Mesh File",
           "*.2dm",
-          Capability::ReadMesh
+          Capability::ReadMesh | Capability::SaveMesh
         )
 {
 }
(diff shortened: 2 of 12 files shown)


mdal/api/mdal.h
```diff
@@ -120,6 +121,10 @@ MDAL_EXPORT const char *MDAL_DR_filters( DriverH driver );
 //! This may effectively load whole mesh in-memory for some providers
 //! Caller must free memory with MDAL_CloseMesh() afterwards
 MDAL_EXPORT MeshH MDAL_LoadMesh( const char *meshFile );
+
+MDAL_EXPORT void MDAL_SaveMesh( MeshH mesh, const char *meshFile );
```
PeterPetrik: please add documentation. Specially: is this saving only mesh file or also datasets? Also maybe worth to add a function to test driver capability to do this operation (see MDAL_DR_writeDatasetsCapability)

mdal/frmts/mdal_2dm.cpp
```diff
@@ -286,3 +286,63 @@ std::unique_ptr<MDAL::Mesh> MDAL::Driver2dm::load( const std::string &meshFile,
...
+  double vertex[3];
+  for ( size_t i = 0; i < mesh->verticesCount(); ++i )
+  {
+    vertexIterator->next( 1, vertex );
+    line = "ND ";
+    line.append( std::to_string( i + 1 ) );
+    for ( size_t j = 0; j < 3; ++j )
+    {
+      line.append( " " );
+      line.append( std::to_string( vertex[j] ) );
```
PeterPetrik: since it is double, we may not write all digits, but only first few decimal places ... maybe add function to   `bool toBool( const std::string &str );` from mdal_utils.hpp?

mdal/frmts/mdal_2dm.cpp
```diff
@@ -286,3 +286,63 @@ std::unique_ptr<MDAL::Mesh> MDAL::Driver2dm::load( const std::string &meshFile,
...
+    }
+    file << line << std::endl;
+  }
+
+  //write faces
+  std::unique_ptr<MDAL::MeshFaceIterator> faceIterator = mesh->readFaces();
+  for ( size_t i = 0; i < mesh->facesCount(); ++i )
+  {
+    int faceOffsets[1];
+    int vertexIndices[4]; //max 4 vertices for a face
```
PeterPetrik: face can have even more than 4 vertices, check function https://github.com/lutraconsulting/MDAL/blob/b8f79d8e4969b1cfc8641257f4bdba28d2c1937a/mdal/api/mdal.h#L141 `MDAL_M_faceVerticesMaximumCount` .If 2dm driver can only store triangles or quads, best to return error code here.

mdal/frmts/mdal_driver.hpp
```diff
@@ -44,6 +44,9 @@ namespace MDAL
       // loads datasets
       virtual void load( const std::string &uri, Mesh *mesh, MDAL_Status *status );
 
+      // save mesh
+      virtual void save( const std::string &uri, MDAL_Status *status, Mesh *mesh );
```
PeterPetrik: can you shuffle arguments so they are in the same order as load function pleasE?

tests/mdal_testutils.hpp
```diff
@@ -8,6 +8,7 @@
 
 #include <string>
 #include <vector>
+#include <cstdio>
```
PeterPetrik: needed?

tests/test_2dm.cpp
```diff
@@ -270,6 +271,98 @@ TEST( Mesh2DMTest, Basement3CellElevationTest )
...
+  std::string pathSource = test_file( "/2dm/quad_and_triangle.2dm" );
+  MeshH meshToSave = MDAL_LoadMesh( pathSource.c_str() );
+
+  //save the mesh
+  std::string fileNameToSave = tmp_file( "/quad_and_triangle_saveTest.2dm" );
+  MDAL_SaveMesh( meshToSave, fileNameToSave.c_str() );
+
+  MDAL_CloseMesh( meshToSave );
+
+  //open the saved mesh and test it (same as QuadAndTriangleFile test)
```
PeterPetrik: do not copy paste (DRY) ... either create common function to test both files, or just compare quad_and_triangle.2dm and quad_and_triangle_saveTest.2dm as text files?

tests/test_2dm.cpp
```diff
@@ -270,6 +271,98 @@ TEST( Mesh2DMTest, Basement3CellElevationTest )
   MDAL_CloseMesh( m );
 }
 
+
+
+TEST( Mesh2DMTest, SaveMeshToFile )
```
PeterPetrik: can you please add test where you want to store file with > 4 vertices per face?

mdal/api/mdal.h
vcloarec: done.
Right now, I think it is the responsibility of the driver to handle the capability to save datasets.

mdal/frmts/mdal_2dm.cpp
vcloarec: I am not sure to understand the link between double digits and `bool`.

mdal/frmts/mdal_2dm.cpp
vcloarec: In the last commit, I have added a method in the MDAL::Driver class to return the maximum vertices per face the driver handles. Default is big, and Diver2DM return 4.
So, now we can test if the mesh is compatible with the driver before saving.

mdal/frmts/mdal_driver.hpp
vcloarec: done

tests/mdal_testutils.hpp
vcloarec: no...

tests/test_2dm.cpp
vcloarec: done

tests/test_2dm.cpp
vcloarec: Do you know if there is some file available with > 4 vertices per face in the test data ?

mdal/api/mdal.h
```diff
@@ -122,11 +125,13 @@ MDAL_EXPORT const char *MDAL_DR_filters( DriverH driver );
 //! Caller must free memory with MDAL_CloseMesh() afterwards
 MDAL_EXPORT MeshH MDAL_LoadMesh( const char *meshFile );
 
-MDAL_EXPORT void MDAL_SaveMesh( MeshH mesh, const char *meshFile );
-
-
 //! Closes mesh, frees the memory
 MDAL_EXPORT void MDAL_CloseMesh( MeshH mesh );
+
+//! Save mesh on a file whith the specified driver. On error see MDAL_LastStatus for error type.
```
PeterPetrik: Saves (3rd person). typo in "whith"

mdal/api/mdal.h
```diff
@@ -122,11 +125,13 @@ MDAL_EXPORT const char *MDAL_DR_filters( DriverH driver );
...
+//! Save the datasets if the driver has the capability
```
PeterPetrik: I think the datasets are not saved in your implementation?

mdal/frmts/mdal_driver.hpp
```diff
@@ -39,13 +40,17 @@ namespace MDAL
 
       virtual bool canRead( const std::string &uri ) = 0;
 
+      //return the maximum vertices per mesh for this driver, default value is big
```
PeterPetrik: Returns (3rd person), missing ! and space. remove "default is big".  I would keep the function as pure virtual and let all drivers implement it (the mesh readers already do it). Or I would keep it as negative number (since some driver even do not have ability to read/write mesh at all)

vcloarec: > since some driver even do not have ability to read/write mesh at all

I think it is better to return a negative number for default and for or avoiding to implement this method for all the drivers

mdal/api/mdal.h
vcloarec: Does 2DM format support datasets ?

PeterPetrik: no, 2dm is mesh-only . datasets is XMDF or dat files for TUFLOW

vcloarec: So, as I have only implemented the save functionality for 2DM, there is no need  implementation of datasets saving for 2DM format. no ?

PeterPetrik: Mesh (the MeshH structure) is a) mesh definition (vertices and faces) and b) datasets ... so ... If I have loaded mesh with datasets, how do I store it...

vcloarec: If the user wants to store the datasets, he has to choose a format that support the datasets. If 2DM doesn't support datasets, he has to choose another format ... As only the 2DM saving capability is implement right now, it is not possible for the moment.
The aim of this PR is to implement 2D format ...

PeterPetrik: I am fine with implementation. Just please add to docs that it does not saves datasets, only mesh definition/frame

vcloarec: I purpose :
```
//! Saves mesh on a file with the specified driver. On error see MDAL_LastStatus for error type.
//! Saves the datasets if the driver has the capability (as, right now, only the 2DM format is implemented for saving,
//! it does not save the datasets except the terrain elevation of the mesh (Z value of vertices)).
```

PeterPetrik: I think we are not on the same page here. I propose that this function only and ever saves only mesh frame. And if you want to save datasets, you call different function to save datasets. This is because you can have a mesh that consists of 2dm frame, but XMDF datasets or any other driver combinations...

vcloarec: Ok, I understand.
But, don't you think it is interesting to have the possibility to save the mesh structure and the datasets in the same file if a driver can't support this ? 

But I don't know enough each format to know if such format is present .... 

In all cases, you are right, implement a function who saves only the mesh structure is interesting. The function who saves mesh+datasets could be a distinct function and could be implemented later.
```

//! Saves mesh (only mesh structure) on a file with the specified driver. On error see MDAL_LastStatus for error type.

```

mdal/frmts/mdal_2dm.cpp
vcloarec: You are right to tell me to be watchful with transform float to string.
This `std::to_string(double)` has only a precision with only 6 digits after the digital point ... With projection coordinates it is ok, but with long/lat, it is a problem because longitudinal length of 1e-5° is equivalent to 1 m when latitude is 0°. So only 6 digits after the digital point is not enough...

But with projected coordinates, more than 2 or 3 numbers after digital point is too much...

In the other hand, for Z value or datasets value, I think it is more pertinent to consider significant digits.

So I purpose to introduce two utils function to handle precision.

tests/test_2dm.cpp
vcloarec: I have added a test with a fake MemoryMesh with maximum vertices per face equal to 5.

mdal/frmts/mdal_2dm.cpp
PeterPetrik: https://xkcd.com/2170/

tests/test_2dm.cpp
PeterPetrik: MemoryMesh is not exported from the mdal library, so you cannot use it (it is implementation detail). if you can just remove this test for the time being

vcloarec: done

PeterPetrik:
thanks, good job

PR #141, opened 2019-08-05
[anuga] Fix crash when reading of files with variables named XXX_c

### LINKED ISSUE #137: Not all SWW files are supported
See https://github.com/qgis/QGIS/issues/30828

wonder-sk:
These are obviously volume-based variables - with value for each volume instead of value for each vertex

Fixes #137


wonder-sk:
Tests needed adjusting - fixed in https://github.com/lutraconsulting/MDAL/commit/6a8a1484a6a5e2cb102e2aa614acca7fd0be6e2e

PR #145, opened 2019-08-29
Fix issue with CF format when no time is present

vcloarec:
Hi,
I tried to read my own UGRID format mesh file with MDAL and there was an issue because no time dimension is present.
This PR fixes the issue.
I have added a test with my own file.

Fix another issue to make MDAL more UGRID compliant (the time dimension is not necessary the first dimension for variable).


PeterPetrik:
sorry for delay, we were attending FOSS4G. Good job.

vcloarec:
no problem @PeterPetrik . I have seen you have presented MDAL at the FOSS4G. Is there a video online available ?

saberraz:
@vcloarec here it is:
https://media.ccc.de/v/bucharest-109-mesh-gis-data-beyond-raster-and-vector

PR #157, opened 2019-09-15
UGRID format : searching the CRS variable name in the X variable

vcloarec:
Hello,

The coordinate system variable name was searched in the Z node variable name. As this variable is no required,  it will be better to search in the X variable name (or Y variable name).

I have added code in the UGRID tests for CRS.


PeterPetrik:
looks good , thanks

Issue #152, opened 2019-09-05
Read and expose reference time in API


PeterPetrik:

PeterPetrik:
implemented `MDAL_EXPORT const char *MDAL_G_referenceTime( DatasetGroupH group );`

PR #161, opened 2019-10-08
#6 Time unit and reference time readout for DAT and XMDF

MUrbanLutraConsulting:


mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -226,9 +227,30 @@ void MDAL::DriverAsciiDat::loadNewFormat( std::ifstream &in,
       if ( quoteIdx1 != std::string::npos && quoteIdx2 != std::string::npos )
         group->setName( line.substr( quoteIdx1 + 1, quoteIdx2 - quoteIdx1 - 1 ) );
     }
+    else if ( cardType == "RT_JULIAN" && items.size() >= 2 )
+    {
+      size_t quoteIdx1 = line.find( " " );
+      size_t quoteIdx2 = line.size();
+      if ( quoteIdx1 != std::string::npos && quoteIdx2 != std::string::npos )
+        referenceTime = ( line.substr( quoteIdx1 + 1, quoteIdx2 - quoteIdx1 - 1 ) );
```
PeterPetrik: isn't it just `referenceTime = items[ 1 ];`?

mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -226,9 +227,30 @@ void MDAL::DriverAsciiDat::loadNewFormat( std::ifstream &in,
...
+      if ( !group )
+      {
+        debug( "TIMEUNITS card for no active dataset!" );
+        EXIT_WITH_ERROR( MDAL_Status::Err_UnknownFormat );
+      }
+
+      size_t quoteIdx1 = line.find( " " );
+      size_t quoteIdx2 = line.size();
+      if ( quoteIdx1 != std::string::npos && quoteIdx2 != std::string::npos )
+        group->setMetadata( "TIMEUNITS", line.substr( quoteIdx1 + 1, quoteIdx2 - quoteIdx1 - 1 ) );
```
PeterPetrik: isn't just just 
```std::string timeunits = items[ 1 ];```?

File: mdal/frmts/mdal_ascii_dat.cpp [hunk 3]
```diff
@@ -226,9 +227,30 @@ void MDAL::DriverAsciiDat::loadNewFormat( std::ifstream &in,
...
+    }
     else if ( cardType == "TS" && items.size() >= 3 )
     {
       double t = toDouble( items[2] );
+      t = convertTimeDataToHours( t, group->getMetadata( "TIMEUNITS" ) );
```
PeterPetrik: give it a default value of seconds, so in case there is no TIMEUNITS in the file it still gets the same result as before

mdal/frmts/mdal_binary_dat.cpp
```diff
@@ -245,6 +248,44 @@ void MDAL::DriverBinaryDat::load( const std::string &datFile, MDAL::Mesh *mesh,
...
+      case CT_RT_JULIAN:
+        // Reference time
+        if ( readIStat( in, sflg, &istat ) )
+          EXIT_WITH_ERROR( MDAL_Status::Err_UnknownFormat );
+
+        if ( read( in, reinterpret_cast< char * >( &time ), 8 ) )
+          EXIT_WITH_ERROR( MDAL_Status::Err_UnknownFormat );
+
+        referenceTime = static_cast<double>( time );
+        group->setReferenceTime( std::to_string( referenceTime ) );
```
PeterPetrik: I would put the same string as is in ASCII, so `"JULIAN`" + std::to_string( referenceTime )` or something like that

tests/test_xmdf.cpp
```diff
@@ -206,7 +206,7 @@ TEST( MeshXmdfTest, CustomGroupsDataset )
   ASSERT_NE( g, nullptr );
 
   int meta_count = MDAL_G_metadataCount( g );
-  ASSERT_EQ( 1, meta_count );
+  ASSERT_EQ( 2, meta_count );
 
   const char *name = MDAL_G_name( g );
   EXPECT_EQ( std::string( "Vector Velocity" ), std::string( name ) );
```
PeterPetrik: - can you please add test to check the metadata for timeunits and reference time for all ascii/binary/xmdf in case we have such testcase?

... so it means that we have no testcase for some other timeunits than defaults?, since you did not have to change any time ASSERT in any test. In that case I suggest to manually create a testcase (by let say copy-pasting some small ascii dat file) with other types of timeunit

PeterPetrik:
related to #152

Issue #162, opened 2019-10-09
Add support for reading arcinfo adf tin layers


nyalldawson:
GDAL/OGR has support for reading arcinfo binary vector and raster layers, but not adf tin layers. Seems like a perfect fit for mdal...

vcloarec:
As a former user of Esri TIN, I am very tempted to do the work... 
Have you got a recent Esri TIN to provide? As I don't use those format for years, I don't have one.

nyalldawson:
I can provide a few, but not ones which could be publicly shared, sorry. Want me to email them?

vcloarec:
yes, it would be great for first trials. But not too big please.
Then we need some which can be publicly for tests.

nyalldawson:
What's your email @vcloarec ?

vcloarec:
Sorry, you can use this one : vcloarec@reos.site

vcloarec:
On the good way : can read the (x,y,z) of the node.

saberraz:
Good overview of the format and specifications, with link to other resources:
https://en.wikipedia.org/wiki/Esri_TIN

PeterPetrik:
I am wondering if this format is ONLY mesh definition or if it has any datasets? If there are no datasets defined on mesh, I am not sure if it not rather part of GDAL/OGR instead of mesh

Wondering where we can get some examples @saberraz do we write to ESRI to provide some or are there any freely available?

PeterPetrik:
also @nyalldawson can you please share some test data with us on info@lutraconsulting.co.uk please?

wonder-sk:
> I am wondering if this format is ONLY mesh definition or if it has any datasets? If there are no datasets defined on mesh, I am not sure if it not rather part of GDAL/OGR instead of mesh

Even if it is only mesh without any extra datasets, it still makes sense to have native support in MDAL for such TINs - GDAL/OGR would need to convert the triangles to a regular grid, loosing some data during the conversion...

saberraz:
An example TIN here:
https://github.com/usgin/Workshops/tree/master/GeothermalResourcesCouncil/MeetingProjects/Arc3D/Bangor_3D/top

No licence according to @ccaudill

PeterPetrik:
no problem, but the limitation will be the `int` (https://github.com/lutraconsulting/MDAL/blob/master/mdal/api/mdal.h#L222) since we internally need to index vertices/faces.

nyalldawson:
[...]

PR #164, opened 2019-10-11
Add support for reading arcinfo adf tin layers #162

vcloarec:
First Draft, the mesh seems to be loaded. Need tests to verify if the geometry is ok.


saberraz:
@vcloarec nice. 
I tried it with the attached mesh.

```
mdalinfo tnz.adf
Mesh loaded: OK
  Driver: ESRI_TIN
  Vertex count: 13
  Face count: 19
  Projection: undefined
Datasets loaded: OK
  Groups count: 1
  Terrain Elevation
```

But when I try to open it in QGIS, I get core dump.
[top.zip](https://github.com/lutraconsulting/MDAL/files/3715702/top.zip)

PS: we can probably use the attached TIN for test and credit http://lab.usgin.org

mdal/frmts/mdal_esri_tin.cpp
```diff
@@ -0,0 +1,179 @@
+/*
+ MDAL - Mesh Data Abstraction Library (MIT License)
+ Copyright (C) 2019 Vincent Cloarec (vcloarec at reos dot site)
+*/
+
+#include "mdal_esri_tin.hpp"
+
+//from https://stackoverflow.com/questions/41012414/convert-double-value-from-little-endian-to-big-endian
+template <typename T>
```
PeterPetrik: I think we have a function for this in https://github.com/lutraconsulting/MDAL/blob/master/mdal/frmts/mdal_selafin.cpp#L145

maybe we can just move that function to MDAL utils and use it here too?

mdal/frmts/mdal_esri_tin.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+    std::ifstream zIn( zFile( uri ), std::ifstream::in | std::ifstream::binary );
+    if ( ! zIn.is_open() )
+        return false;
+
+    std::ifstream faceIn( faceFile( uri ), std::ifstream::in | std::ifstream::binary );
+    if ( ! faceIn.is_open() )
+        return false;
+
+    //check if the file sizes of the nodes are coherent ?
```
PeterPetrik: i think it does not need to be... what you can check is first line of some file if they have some standard header/first line

mdal/frmts/mdal_esri_tin.cpp
```diff
@@ -0,0 +1,179 @@
...
+}
+
+std::string MDAL::DriverEsriTin::faceFile(const std::string &uri) const
+{
+    return pathJoin( dirName( uri ), "tnod.adf" );
+}
+
+void MDAL::DriverEsriTin::populateVertices( const std::string &uri, MDAL::Vertices &vertices )
+{
+    //firt, read the (x,y) coordinates
```
PeterPetrik: typo: first

mdal/frmts/mdal_esri_tin.cpp
```diff
@@ -0,0 +1,179 @@
...
+    std::ifstream inXY( xyFile( uri ), std::ifstream::in | std::ifstream::binary );
+
+    if ( ! inXY.is_open() )
+        throw MDAL_Status::Err_FileNotFound;
+
+    inXY.ignore( 64 ); //ignore the 64 first bytes
```
PeterPetrik: this is probably something to function canRead(). aren't those 64bytes some common string?

mdal/frmts/mdal_esri_tin.cpp
```diff
@@ -0,0 +1,179 @@
...
+    //bool readingOk=(!inXY.eof());
```
PeterPetrik: remove commented out code

mdal/frmts/mdal_esri_tin.cpp
```diff
@@ -0,0 +1,179 @@
...
+    while ( true )
+    {
+        Vertex vert;
+
+        if ( !readAndChangeEndianness( vert.x, inXY ) )
```
PeterPetrik: this is hard to read. I would extract the in.read to this loop and handle reading double in another routine.

mdal/frmts/mdal_esri_tin.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+    inXY.close();
+
+    //Then, read Z value
+    std::ifstream inZ( zFile( uri ), std::ifstream::in | std::ifstream::binary );
+  if ( ! inZ.is_open() )
+    throw MDAL_Status::Err_FileNotFound;
+
+  inZ.ignore( 16 ); //ignore the 16 first bytes
+  //readingOk=(!inZ.eof());
```
PeterPetrik: remove

mdal/frmts/mdal_esri_tin.cpp
```diff
@@ -0,0 +1,179 @@
...
+      f.push_back( static_cast<size_t>( ind ) ); // -5 to ignore "superpoints"
+    }
+
+    if ( f.size() == 0 ) //that's mean this is the end of the file
+      break;
+
+    if ( f.size() < 3 ) //that's mean the face is not complete
+      throw MDAL_Status::Err_UnknownFormat;
+
+    //adjust the indexes and exclude face with "superpoints
```
PeterPetrik: missing "

what is superpoint? can you add description here?

mdal/frmts/mdal_esri_tin.hpp
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * 1 2 3 4 5 6 7 8 / 1 2 3 4 5 6 7 8
+   *         ?       /    (ymax+ymin)/2   //superpoint
+   *  (xmax+xmin)/2  /         ?          //superpoint
+   *        ?        /    (ymax+ymin)/2  //superpoint
+   *  (xmax+xmin)/2  /         ?         //superpoint
+   *
+   *
+   *
+   * other bytess: X,Y : for each value 8 bits : double with big endianness
```
PeterPetrik: typo

mdal/frmts/mdal_esri_tin.hpp
```diff
@@ -0,0 +1,79 @@
...
+   *
+   *
+   *  tnz.adf : z value of each vertices
+   * first 16 bits seems to be "informations:"superpoint" (infinity point to North, South East and West)
+   *
+   * other bits : Z : for each value 4 bites : float with big endiannes
+   *
+   *
+  */
+
```
PeterPetrik: remove extra space/newlines all over the file

mdal/mdal_utils.cpp
```diff
@@ -487,7 +488,7 @@ void MDAL::addBedElevationDatasetGroup( MDAL::Mesh *mesh, const Vertices &vertic
                                           mesh->driverName(),
                                           mesh,
                                           mesh->uri(),
-                                          "Bed Elevation"
+                                          "Terrain Elevation"
```
PeterPetrik: Please keep Bed elevation. We are using the term for this for 10 years...

mdal/mdal_utils.cpp
```diff
@@ -505,6 +506,12 @@ void MDAL::addBedElevationDatasetGroup( MDAL::Mesh *mesh, const Vertices &vertic
   mesh->datasetGroups.push_back( group );
 }
 
+void MDAL::addBedElevationDatasetGroup( MDAL::Mesh *mesh, const Vertices &vertices )
```
PeterPetrik: I would just use Bed elevation, not Terrain elevation . It is standard

tests/CMakeLists.txt
```diff
@@ -72,7 +72,7 @@ SET(TESTS
     test_binary_dat.cpp
     test_mdal_utils.cpp
     test_selafin.cpp
-)
+    test_esri_tin.cpp)
```
PeterPetrik: missing newline before )

tests/test_esri_tin.cpp
```diff
@@ -0,0 +1,38 @@
...
+  MeshH m = MDAL_LoadMesh( path.c_str() );
+
+  ASSERT_TRUE( nullptr != m );
+
+  int v_count = MDAL_M_vertexCount( m );
+  EXPECT_EQ( v_count, 13 );
+
+  int f_count = MDAL_M_faceCount( m );
+  EXPECT_EQ( f_count, 19 );
+
```
PeterPetrik: please add test for few faces to see if the vertices are correctly assigned. Also add check for bed elevation dataset

nyalldawson:
Wow, great work.... That was super fast!!

vcloarec:
@saberraz 
The issue is may be fix, it was bad indexes for faces

vcloarec:
@PeterPetrik , I try to look at all your comments may be later today.

mdal/frmts/mdal_esri_tin.cpp
vcloarec: There isn't more information that the position of the points and the indexes of the face. The unused first bytes (64 bytes for tnxy.adf, and 16 bytes for tnz.adf) are used by ArcGIS to store information about 4 virtual points created by ArcGIS for triangulation/generation (https://en.wikipedia.org/wiki/Esri_TIN, see Superpoint). Those points seem to be necessary for ArcGIS but not for MDAL.

So, for me, the only information to know if a file can be read is :
- if it can be opened
- if the size of the tns.adf file (Z value file) is coherent with the size of of tnxy.adf file.
I am not sure of the relevance of checking the size file coherence.

mdal/frmts/mdal_esri_tin.cpp
vcloarec: No, see the other comment

mdal/frmts/mdal_esri_tin.cpp
vcloarec: Sorry, I am not sure to understand this note.

mdal/frmts/mdal_esri_tin.cpp
vcloarec: Description is in the top of `mdal/frmts/mdal_esri_tin.hpp` file but I will try to be more explicit.

mdal/mdal_utils.cpp
vcloarec: For me "Bed Elevation" is specific for river bed elevation. Terrain Elevation could be more generic. As TIN are note only for river, that's why I purpose to keep Bed Elevation for format specific to hydraulic modeling, and maybe we can purpose another name for other format not specific to river.

saberraz:
Nice, my test mesh works now!
test

mdal/frmts/mdal_esri_tin.cpp
PeterPetrik: lets keep it as it is, just remove commented out code please

mdal/frmts/mdal_esri_tin.cpp
PeterPetrik: my point is that reading of the stream should be outside of the function. One function - one task... 

other question was how it behaves if you host system has big endianness too... I was refering to selafin implementation , see https://github.com/lutraconsulting/MDAL/blob/master/mdal/frmts/mdal_selafin.cpp#L145

mdal/frmts/mdal_esri_tin.cpp
PeterPetrik: thanks

mdal/mdal_utils.cpp
PeterPetrik: whatabout keeping the Bed elevation and just change the name in the TIN driver...?

vcloarec:
> Nice, my test mesh works now!
It is not the prettier TIN I've seen :)
The non-public TIN sent by Nyall look better!

mdal/frmts/mdal_esri_tin.cpp
vcloarec: I have added new function in MDAL utils, and I have changed selafin driver and esri tin driver.
Tell me what do you think about ?
But I am not sure that it is a good idea because the number of bytes read in the file depend now of the size of type. Is there a risk that this size of type is different from platform to other ? If yes, it is maybe not a good idea ...

nyalldawson:
I've got some nice simple reference files incoming (freely licensed). Will post a link when I receive them!

vcloarec:
> I've got some nice simple reference files incoming (freely licensed). Will post a link when I receive them!

Is it possible to have TIN files with holes ?

nyalldawson:
I'll ask. Also trying to get some with tagged edges/vertices...

vcloarec:
Hello,
With this last commit, I think MDAL reads ESRI TIN as well as possible.  Holes haven't been tested, but with coming Nyall files, maybe it will be done.
My comments on particular points below :

**Superpoints :**
If the first bytes of xy file and z file are usually used for particular points (called "superpoints" here) that are not used by MDAL, it seems it is not always the case. The file "thul.adf" contains the indexes of the superpoints that can be no the first. The code has been adapted to take into account superpoints that are not at the 
beginning. But as we don't have a file with superpoints not at the beginning, it is not tested in such a situation.

**Mask :**
ArcGIS uses the mask to make invisible faces, those which are outside the outer boundaries, in holes or linked with superpoints. In this PR, MDAL simply used the mask file to ignore those faces.

**Name of the TIN :**
ESRI TINs are saved in a folder that the user gives when he creates it in ArcGis. All the different files in this folder always have the same names and have all the extension *.adf. As MDAL and QGIS used a filename to load a mesh, the trick is to retrieve the folder with any files with *.adf extension is provided to MDAL. So, in QGIS,  selecting any *.adf files permits to load the TIN. The bad consequence is the name of the mesh is the name of the selected file, not the name of the folder...
I have implemented a method in the ESRI TIN driver to retrieve the name of the TIN. But now, it is not used because it seems not trivial for QGIS to obtain this name...

**Endianness :**
The endianness is handled by a new function in MDAL utils. This template function permits to read the value of any type in a binary file. An option provides the possibility to change the endianness of the value if needed.
Depending on the platform, if the type size is not compatible with the same type in the file, the caller keeps the responsibility to handle with this.
For Selafin driver, I have changed the `read_double()` method to see how this method could be improved. But, as some other methods could have to be changed too, I purpose to restore the initial code of this method. The Selafin driver could be improved in another commit/PR if the new template function is pertinent.

mdal/frmts/mdal_esri_tin.cpp
```diff
@@ -0,0 +1,300 @@
...
+std::string MDAL::DriverEsriTin::getCrsWkt( const std::string &uri ) const
+{
+  std::ifstream inCRS( crsFile( uri ), std::ifstream::in );
+  if ( ! inCRS.is_open() )
+    return std::string();
+
+  std::string crsWkt;
+  std::getline( inCRS, crsWkt );
+
+  if ( crsWkt == "{B286C06B-0879-11D2-AACA-00C04FA33C20}" )
```
PeterPetrik: I do not understand this {B28....}? 

Wouldn't it be more sensible to read projection from *.prj file? (3D_Boreholes.prj in the test case?)

nyalldawson: It's a "magic sequence" used in ESRI to indicate a "not set" or unknown CRS. (It's actually the com class id of the esri UnknownCoordinateSystem class).

PeterPetrik: do you know if this file contains actually WKT or it is always some magic sequence of class ids?

vcloarec: I don't know why the *.prj file is here. With other TIN, there is not this file but there is always a prj.adf file. The prj.adf file contains WKT.

PeterPetrik: ok no problem
, can you add comment on the line?

vcloarec: Done

PeterPetrik:
thanks a lot, merging!

nyalldawson:
Awesome work -- this is very exciting!

@PeterPetrik any chance of a mdal version bump prior to QGIS 3.10.0?

nyalldawson:
@vcloarec here's a zipped archive of a bunch of freely-usable tin reference files

[tins.zip](https://github.com/lutraconsulting/MDAL/files/3724126/tins.zip)

It includes the original dems, the arcgis tins, and the results of exporting the tins to polygon layers (in the gdb file). 

Hope it helps verify the holes/islands cases!

vcloarec:
Tanks @nyalldawson , I quickly tried to load the mesh in QGIS, and all have been loaded, with holes and islands. I will add this TIN to the test folder and add tests, but later, I don't have time anymore...
I am very happy this enhancement seems to be in the next release of QGIS!

nyalldawson:
@vcloarec 

For completeness, the attached updates to the tins adds an example with tagged vertices. That should cover all the functionality possible in this format:

[dem (2).zip](https://github.com/lutraconsulting/MDAL/files/3737407/dem.2.zip)

PeterPetrik:
@nyalldawson @vcloarec any possibility you have time to add those to test suite?

vcloarec:
> 
> 
> @nyalldawson @vcloarec any possibility you have time to add those to test suite?

I don't have a lot of time this week for that. I will try next week.

vcloarec:
Hi,
I started to add the test with the TINs give by @nyalldawson. 
First test, no problem, the vertices count and the faces count read by MDAL are the same than in the gdb files.
When testing the dem with hole, there is a little difference with vertices count between MDAL and the gdb file. Two of those extra vertices are in blue in the picture below (the underlayer is the raster TIN, the source ?)
image
 
I think, during any of the conversions, faces have gone away and the vertices stay alone... Maybe, this is due of the border of the raster TIN.

So, in the ESRI TIN files, it seems there are vertices without faces ... 
I purpose to implement a process to remove those isolated vertices, but it will slow down the loading...
What do you think about ?

nyalldawson:
@vcloarec I can't work out quite where in the tin you are looking -- can you show the whole extent?

nyalldawson:
@vcloarec disregard that, sorry!

nyalldawson:
@vcloarec 
Here's what I see in arcmap:

image

So those extra vertices are attached only to edges with a type of "Outside edge" (3). I gather you are currently stripping these away during the conversion?

vcloarec:
@nyalldawson, yes, outside faces (so implicitly associated edges, as MDAL doesn't store directly the edges) are stripped away.
For my information, what is the relation between the raster DEM (tif) and the TIN ? If the raster DEM is a conversion from the TIN, it is weird that some pixels are outside the boundary :
Screenshot_20191025_082040


In any cases, even if there is no visual consequences to keep the isolated vertices, I think it is cleaner to strip away this vertices too when loading the mesh.

PR #166, opened 2019-10-16
[Feature] Write support for ASCII_DAT

MarcusUrban:


mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -444,6 +444,77 @@ void MDAL::DriverAsciiDat::readFaceTimestep(
...
+
+        for ( size_t i = 0; i < valuesToRead; ++i )
+        {
+          // Read values flags
+          if ( isScalar )
+              out << dataset->values()[i] << "\n";
+          else
+          {
+              out << dataset->values()[2 * i] << " " << dataset->values()[2 * i + 1 ]  << "\n";
+          }
```
PeterPetrik: why not just `for (size_t i=0; i< dataset->values().size(); ++i)` ... you need to flush all values anyways

mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -444,6 +444,77 @@ void MDAL::DriverAsciiDat::readFaceTimestep(
...
+        out << "TS " << time_index << "        " << std::to_string(dataset->time()) << "\n";
+        size_t valuesToRead = 0;
+        if (isOnVerticies)
+            valuesToRead = nodeCount;
+        else
+            valuesToRead = elemCount;
+
+        for ( size_t i = 0; i < valuesToRead; ++i )
+        {
+            bool active = static_cast<bool>( dataset->active()[i] );
```
PeterPetrik: I think active flag is stored only for datasets defined on vertices. for datasets on faces it is not defined. Or when active flag = false, you just write NaN as face value...

tests/test_ascii_dat.cpp
```diff
@@ -534,6 +534,206 @@ TEST( MeshAsciiDatTest, QuadAndTriangleVertexVectorOldFile )
...
+    ASSERT_EQ( 2, meta_count );
+
+    const char *name = MDAL_G_name( g );
+    EXPECT_EQ( std::string( "vectorGrp" ), std::string( name ) );
+
+    bool scalar = MDAL_G_hasScalarData( g );
+    EXPECT_EQ( false, scalar );
+
+    bool onVertices = MDAL_G_isOnVertices( g );
+    EXPECT_EQ( true, onVertices );
```
PeterPetrik: you are missing test cases for both scalar dataset defined on faces and vector dataset defined on faces

saberraz:
Do we have a ticket for this PR?

PeterPetrik:
> Do we have a ticket for this PR?

no,  but we need to store element results to hec-ras and flo-2d which is considerably more diffuculy. This is a debugging tool for development

mdal/frmts/mdal_ascii_dat.cpp
MarcusUrban: values() is not returning a std::vector<> but double*

mdal/frmts/mdal_ascii_dat.cpp
MarcusUrban: Well it is not mandatory so lets skip the loop completely for face datasets.

mdal/frmts/mdal_ascii_dat.cpp
PeterPetrik: whatabout `dataset->valuesCount()`

mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -444,6 +444,77 @@ void MDAL::DriverAsciiDat::readFaceTimestep(
...
+        out << "RT_JULIAN " << referenceTimeStr << "\n";
+
+    out << "TIMEUNITS " << 0 << "\n";
+
+    for ( size_t time_index = 0; time_index < group->datasets.size(); ++ time_index )
+    {
+        const std::shared_ptr<MDAL::MemoryDataset> dataset
+                    = std::dynamic_pointer_cast<MDAL::MemoryDataset>( group->datasets[time_index] );
+
+        out << "TS " << time_index << "        " << std::to_string(dataset->time()) << "\n";
```
MarcusUrban: This part is wrong as well this is not a timestamp index but a flag saying if active flags are present or not.

mdal/frmts/mdal_driver.cpp
```diff
@@ -87,7 +87,12 @@ void MDAL::Driver::createDataset( MDAL::DatasetGroup *group, double time, const
 {
   std::shared_ptr<MDAL::MemoryDataset> dataset = std::make_shared< MemoryDataset >( group );
   dataset->setTime( time );
-  memcpy( dataset->values(), values, sizeof( double ) * dataset->valuesCount() );
+  size_t count = dataset->valuesCount();
+
+  if ( !group->isScalar() )
+    count *= 2;
```
MarcusUrban: Please check if this is OK, without this only half of the array gets copied for vector data.

mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -444,6 +444,77 @@ void MDAL::DriverAsciiDat::readFaceTimestep(
   group->datasets.push_back( dataset );
 }
 
+bool MDAL::DriverAsciiDat::persist( MDAL::DatasetGroup *group )
+{
+    std::ofstream out( group->uri(), std::ofstream::out );
```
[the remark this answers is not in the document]
MarcusUrban: In the reader part of the code it looks for "_els_ in the file name ( current line 148). Currently when I am writing a file, I simply add _els_ at the end of the file name before the .dat file extension. Which is not really ideal since it gives files looking like this xxxyyy_els_.dat. I am assuming _els_ position is not actually defined and it can be anywhere.

If we can pick, we should probably agree on something better, either adding _els_ into another part of the file name or adding just _els at the end? (xxxyy_els.dat - looks much better),

mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -444,6 +444,101 @@ void MDAL::DriverAsciiDat::readFaceTimestep(
...
+  if ( !MDAL::contains( uri, "_els_" ) && isOnVerticies == false )
+  {
+    // Should contain _els in name but it does not
+    uri.insert( uri.size() - 4, "_els_" );
+  }
+  else if ( !MDAL::contains( uri, "_vertex" ) && isOnVerticies == true )
+  {
+    // Should contain _vertex in name but it does not
+    uri.insert( uri.size() - 4, "_vertex" );
+  }
```
PeterPetrik: vertex is by default, shouldn't be added at all

mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -333,14 +333,14 @@ void MDAL::DriverAsciiDat::readVertexTimestep(
...
-  size_t faceCount = mesh->facesCount();
+  size_t vertexCount = mesh->verticesCount();
 
   std::shared_ptr<MDAL::MemoryDataset> dataset = std::make_shared< MDAL::MemoryDataset >( group.get() );
   dataset->setTime( t );
 
   int *active = dataset->active();
   // only for new format
-  for ( size_t i = 0; i < faceCount; ++i )
+  for ( size_t i = 0; i < vertexCount; ++i )
```
PeterPetrik: the loop was correct before. Active flag is defined on faces, not vertices

mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -444,6 +444,101 @@ void MDAL::DriverAsciiDat::readFaceTimestep(
...
+
+  out << "TIMEUNITS " << 0 << "\n";
+
+  for ( size_t time_index = 0; time_index < group->datasets.size(); ++ time_index )
+  {
+    const std::shared_ptr<MDAL::MemoryDataset> dataset
+      = std::dynamic_pointer_cast<MDAL::MemoryDataset>( group->datasets[time_index] );
+
+
+    out << "TS " << static_cast<bool>( dataset->active() ) << " " << std::to_string( dataset->time() ) << "\n";
```
PeterPetrik: instead of `static_cast<bool>( dataset->active() )` it should be off for datasets on faces and of on datasets on vertices

mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -444,6 +444,101 @@ void MDAL::DriverAsciiDat::readFaceTimestep(
...
+      out << "RT_JULIAN " << referenceTimeStrWords[1] << "\n";
+    else
+      out << "RT_JULIAN " << referenceTimeStr << "\n";
+  }
+
+  out << "TIMEUNITS " << 0 << "\n";
+
+  for ( size_t time_index = 0; time_index < group->datasets.size(); ++ time_index )
+  {
+    const std::shared_ptr<MDAL::MemoryDataset> dataset
```
PeterPetrik: it does not need to be memorydataset. You should use base dataset class API

mdal/frmts/mdal_driver.cpp
PeterPetrik: it is ok

mdal/frmts/mdal_ascii_dat.cpp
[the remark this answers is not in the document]
PeterPetrik: it is ok at the end

mdal/frmts/mdal_ascii_dat.cpp
PeterPetrik: ?

PeterPetrik:
you have some memory leaks in the tests ... maybe forgot to close mesh or something else?

mdal/frmts/mdal_ascii_dat.cpp
```diff
@@ -444,6 +445,96 @@ void MDAL::DriverAsciiDat::readFaceTimestep(
...
+    size_t valuesToRead = 0;
+
+    if ( isOnVerticies )
+    {
+      valuesToRead = nodeCount;
+      // Fill the active data
+      for ( size_t i = 0; i < elemCount; ++i )
+      {
+        int active = dataset->active()[i];
+        out << ( active == 1 ? true : false ) << "\n";
```
PeterPetrik: Does this produce 0 and 1 in the resulting file?

PR #170, opened 2019-10-23
Ugrid writer

MarcusUrban:


mdal/frmts/mdal_netcdf.cpp
```diff
@@ -216,3 +214,38 @@ void NetCDFFile::addDimension( const std::string &name, size_t size )
     throw MDAL_Status::Err_FailToWriteToDisk;
   }
 }
+
+int NetCDFFile::defineVar( const std::string &varName,
+                           int ncType, int dimensionCount, const int *dimensions )
```
MarcusUrban: I will probably replace these two with `int dimensionCount, const int *dimensions`  with `std::vector<int>` ?

PeterPetrik: +1

mdal/frmts/mdal_ugrid.hpp
```diff
@@ -45,6 +46,13 @@ namespace MDAL
       std::string mMesh2dName;
       std::string mMesh1dName;
       std::string nodeZVariableName() const;
+
+      void writeDimensions( MDAL::Mesh *mesh );
+      void writeVariables( MDAL::Mesh *mesh );
+      void writeGlobals( MDAL::Mesh *mesh );
+      // TODO: Could not find what this number should be, more investigation needed!
+      int faceVerticesMaximumCount() const override
```
PeterPetrik: just use https://github.com/lutraconsulting/MDAL/blob/2119ad7786c2e14b9b5259cdefdc0f08bcb72a24/mdal/mdal_data_model.hpp#L184

mdal/mdal_utils.hpp
```diff
@@ -110,6 +110,7 @@ namespace MDAL
   // time
   //! Returns a delimiter to get time in hours
   MDAL_TEST_EXPORT double parseTimeUnits( const std::string &units );
+  std::string getCurrentTimeStamp();
```
PeterPetrik: add comment

mdal/frmts/mdal_cf.cpp
```diff
@@ -401,7 +406,8 @@ std::unique_ptr< MDAL::Mesh > MDAL::DriverCF::load( const std::string &fileName,
     );
     mesh->faces = faces;
     mesh->vertices = vertices;
-
+    size_t faceVerticesMaximumCount = mDimensions.size( mDimensions.MaxVerticesInFace );
+    mesh->setFaceVerticesMaximumCount( faceVerticesMaximumCount );
```
PeterPetrik: not sure why it is needed as it is already set few lines above in https://github.com/lutraconsulting/MDAL/blob/master/mdal/frmts/mdal_cf.cpp#L397

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -401,3 +403,133 @@ void MDAL::DriverUgrid::parse2VariablesFromAttribute( const std::string &name, c
...
+
+  // Nodes Z coordinate
+  int mesh2dNodeZId = mNcFile.defineVar( "mesh2d_node_z", NC_DOUBLE, 1, &dimNodeCountId );
+  mNcFile.putAttrStr( mesh2dNodeZId, "mesh", "mesh2d" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "location", "node" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "coordinates", "mesh2d_node_x mesh2d_node_y" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "standard_name", "altitude" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "long_name", "z-coordinate of mesh nodes" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "units", "m" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "grid_mapping", "projected_coordinate_system" );
```
PeterPetrik: you need to store CRS in this array if you aldready defined it, see https://github.com/lutraconsulting/MDAL/blob/master/mdal/mdal_data_model.hpp#L183

but only if it is not-null. (there are diferent options, you probably need to parse back EPSG code if there is one, or store whole proj-string in case it is stored in that way)

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -401,3 +403,133 @@ void MDAL::DriverUgrid::parse2VariablesFromAttribute( const std::string &name, c
...
+  mNcFile.putAttrStr( mesh2dNodeYId, "location", "node" );
+
+  // Nodes Z coordinate
+  int mesh2dNodeZId = mNcFile.defineVar( "mesh2d_node_z", NC_DOUBLE, 1, &dimNodeCountId );
+  mNcFile.putAttrStr( mesh2dNodeZId, "mesh", "mesh2d" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "location", "node" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "coordinates", "mesh2d_node_x mesh2d_node_y" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "standard_name", "altitude" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "long_name", "z-coordinate of mesh nodes" );
+  mNcFile.putAttrStr( mesh2dNodeZId, "units", "m" );
```
PeterPetrik: doesn't need to be meteres, depends on CRS. Is it required in the CF/Ugrid standart to have units?

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -401,3 +403,133 @@ void MDAL::DriverUgrid::parse2VariablesFromAttribute( const std::string &name, c
...
+  int mesh2dNodeXId = mNcFile.defineVar( "mesh2d_node_x", NC_DOUBLE, 1, &dimNodeCountId );
+  mNcFile.putAttrStr( mesh2dNodeXId, "units", "m" );
+  mNcFile.putAttrStr( mesh2dNodeXId, "standard_name", "projection_x_coordinate" );
+  mNcFile.putAttrStr( mesh2dNodeXId, "long_name", "x-coordinate of mesh nodes" );
+  mNcFile.putAttrStr( mesh2dNodeXId, "mesh", "mesh2d" );
+  mNcFile.putAttrStr( mesh2dNodeXId, "location", "node" );
+
+  // Nodes Y coordinate
+  int mesh2dNodeYId = mNcFile.defineVar( "mesh2d_node_y", NC_DOUBLE, 1, &dimNodeCountId );
+  mNcFile.putAttrStr( mesh2dNodeYId, "units", "m" );
```
PeterPetrik: doesn't need to be meteres, depends on CRS. Is it required in the CF/Ugrid standart to have units?

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -401,3 +403,133 @@ void MDAL::DriverUgrid::parse2VariablesFromAttribute( const std::string &name, c
...
+  mNcFile.putAttrInt( mesh2FaceNodesId, "_FillValue", -999 );
+
+  // Turning off define mode - allows data write
+  nc_enddef( mNcFile.handle() );
+
+  // Write vertices
+  std::unique_ptr<MDAL::MeshVertexIterator> vertexIterator = mesh->readVertices();
+
+  const size_t verticesCoordCount = mesh->verticesCount() * 3;
+  double *verticesCoordinates = new double[verticesCoordCount];
```
PeterPetrik: do not use raw pointers/vectors. if mNcFile.putDataDouble raises exception, you got memory leak

also this would not work for meshes that do not fit into memory. see https://github.com/qgis/QGIS/blob/b2e712149312390b483cfb3c9be4c63e875694f0/src/providers/mdal/qgsmdalprovider.cpp#L89 how to iterate with maximum 1000 items at the same time.

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -401,3 +403,133 @@ void MDAL::DriverUgrid::parse2VariablesFromAttribute( const std::string &name, c
...
+  }
+
+  delete[] verticesCoordinates;
+
+  // Write faces
+  std::unique_ptr<MDAL::MeshFaceIterator> faceIterator = mesh->readFaces();
+  const size_t faceVerticesMax = mesh->faceVerticesMaximumCount();
+
+  int faceOffsets[1];
+  int *vertexIndices = new int[faceVerticesMax];
```
PeterPetrik: do not use raw vectors. Also do not iterate one by one, it is not speed-effective. see https://github.com/qgis/QGIS/blob/b2e712149312390b483cfb3c9be4c63e875694f0/src/providers/mdal/qgsmdalprovider.cpp#L116 how to iterate in chunks

mdal/mdal_utils.cpp
```diff
@@ -357,6 +358,13 @@ double MDAL::parseTimeUnits( const std::string &units )
   return divBy;
 }
 
+std::string MDAL::getCurrentTimeStamp()
+{
+  std::time_t now = time( nullptr );
+  char *dateTime = std::ctime( &now );
```
PeterPetrik: shouldn't we use https://en.cppreference.com/w/cpp/chrono/c/strftime instead?

mdal/mdal_data_model.hpp
```diff
@@ -167,6 +167,7 @@ namespace MDAL
       void setSourceCrs( const std::string &str );
       void setSourceCrsFromWKT( const std::string &wkt );
       void setSourceCrsFromEPSG( int code );
+      void setFaceVerticesMaximumCount( size_t value );
```
PeterPetrik: i do not think it is needed. maximum count is set in constructor of the mesh

mdal/frmts/mdal_ugrid.hpp
```diff
@@ -45,6 +46,12 @@ namespace MDAL
       std::string mMesh2dName;
       std::string mMesh1dName;
       std::string nodeZVariableName() const;
+
+      void writeDimensions( MDAL::Mesh *mesh );
+      void writeVariables( MDAL::Mesh *mesh );
+      void writeGlobals();
+      int faceVerticesMaximumCount() const override
+      { return CFDimensions::Type::MaxVerticesInFace; }
```
PeterPetrik: this is just emum value for CFDimensions::Type::MaxVerticesInFace (so maybe 3 or 4 or whatever is the index. of this enum option in the enum). You want probably mDimenstions.size(CFDimensions::Type::MaxVerticesInFace); nevertheless, why the base function from `MDAL::Mesh` is not good enough?

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -401,3 +403,133 @@ void MDAL::DriverUgrid::parse2VariablesFromAttribute( const std::string &name, c
...
+
+  delete[] vertexIndices;
+
+  // Turning on define mode
+  nc_redef( mNcFile.handle() );
+}
+
+void MDAL::DriverUgrid::writeGlobals()
+{
+  mNcFile.putAttrStr( NC_GLOBAL, "institution", "Lutra Consulting Limited" );
```
PeterPetrik: please remove the institution line

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -401,3 +403,133 @@ void MDAL::DriverUgrid::parse2VariablesFromAttribute( const std::string &name, c
...
+  mNcFile.putAttrStr( NC_GLOBAL, "source", "MDAL" );
```
PeterPetrik: please add MDAL version from MDAL_version()

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -401,3 +403,133 @@ void MDAL::DriverUgrid::parse2VariablesFromAttribute( const std::string &name, c
...
+  mNcFile.putAttrStr( NC_GLOBAL, "date_created", MDAL::getCurrentTimeStamp() );
+  mNcFile.putAttrStr( NC_GLOBAL, "Conventions", "CF-1.6 UGRID-1.0 / MDAL" );
```
PeterPetrik: remove "/MDAL" we try to strictly follow UGRID-1.0 conventions.

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -401,3 +403,133 @@ void MDAL::DriverUgrid::parse2VariablesFromAttribute( const std::string &name, c
...
+  mNcFile.putAttrDouble( mesh2dNodeZId, "_FillValue", -999.0 );
+
+  // Faces 2D Variable
+  int mesh2FaceNodesId_dimIds [] { dimFaceCountId, dimMaxNodesPerFaceId };
+  int mesh2FaceNodesId = mNcFile.defineVar( "mesh2d_face_nodes", NC_INT, 2, mesh2FaceNodesId_dimIds );
+  mNcFile.putAttrStr( mesh2FaceNodesId, "cf_role", "face_node_connectivity" );
+  mNcFile.putAttrStr( mesh2FaceNodesId, "mesh", "mesh2d" );
+  mNcFile.putAttrStr( mesh2FaceNodesId, "location", "face" );
+  mNcFile.putAttrStr( mesh2FaceNodesId, "long_name", "Mapping from every face to its corner nodes (counterclockwise)" );
+  mNcFile.putAttrInt( mesh2FaceNodesId, "start_index", 1 );
```
PeterPetrik: in MDAL internal arrays it is numbered from 0. So I think you either need to put 0 here or add "1" to every face/vertex index when you store it. (in case I undestand correctly this attribute...)

tools/mdal2mdal.cpp
```diff
@@ -0,0 +1,101 @@
+/*
+ MDAL - Mesh Data Abstraction Library (MIT License)
+ Copyright (C) 2018 Peter Petrik (zilolv at gmail dot com)
```
PeterPetrik: copy-paste error

PR #179, opened 2019-12-09
time handling

vcloarec:
For discussion


mdal/mdal_data_model.hpp
```diff
@@ -38,6 +38,137 @@ namespace MDAL
 
   typedef std::vector< std::pair< std::string, std::string > > Metadata;
 
+  class DateTime;
```
PeterPetrik: just create a new file with date/time routines

mdal/mdal_data_model.hpp
```diff
@@ -38,6 +38,137 @@ namespace MDAL
...
+      static DateTime fromStandartValue( int year, int month, int day, int hours = 0, int minutes = 0, double seconds = 0 );
+
+      //! Create an instance from julian day
+      static DateTime fromJulianDay( double julianDay )
+      {
+        return DateTime( int64_t( julianDay * 24 * 3600 * 1000 ) );
+      }
+
+      //! Create an default instance
+      static DateTime fromDefault()
```
PeterPetrik: why you need extra contructor like this?

mdal/mdal_data_model.hpp
```diff
@@ -38,6 +38,137 @@ namespace MDAL
...
+      {
+        return DateTime();
+      }
+
+      //! Returns a string with the date/time expressed in Greogrian/Julian calendar with ISO8601 format (local time zone)
```
PeterPetrik: I think we do not need to use local time zone for output. MDAL API should return universal gregorian UTC time string

mdal/mdal_data_model.hpp
```diff
@@ -38,6 +38,137 @@ namespace MDAL
...
+      std::string toStandartCalendarISO8601() const;
+
+      //! Returns the Julian day value
+      double toJulianDay() const;
```
PeterPetrik: why we need this?

mdal/mdal_data_model.hpp
```diff
@@ -38,6 +38,137 @@ namespace MDAL
...
+
+      //! Returns the Julain day value expressed with a string
+      std::string toJulianDayString() const
```
PeterPetrik: why we need this?

mdal/mdal_data_model.hpp
```diff
@@ -38,6 +38,137 @@ namespace MDAL
 
   typedef std::vector< std::pair< std::string, std::string > > Metadata;
 
+  class DateTime;
+
+  class Duration
```
PeterPetrik: I am wondering if we shouldn't use Time and DateTime (similar to QTime and QDataTime), but I am happy with Duration too

mdal/mdal_data_model.cpp
```diff
@@ -332,3 +337,293 @@ size_t MDAL::Mesh::faceVerticesMaximumCount() const
...
+}
+
+double MDAL::Duration::value( MDAL::Duration::Unit unit ) const
+{
+  switch ( unit )
+  {
+    case MDAL::Duration::milliseconds:
+      return double( mDuration );
+    case MDAL::Duration::seconds:
+      return double( mDuration ) / 1000 ;
```
PeterPetrik: better to divide/multiply/add/distract by floating point number, add .0 to all numbers everyewhere

mdal/mdal_data_model.cpp
```diff
@@ -332,3 +337,293 @@ size_t MDAL::Mesh::faceVerticesMaximumCount() const
...
+      mDuration = int64_t( duration * 60 * 1000 );
+      break;
+    case MDAL::Duration::hours:
+      mDuration = int64_t( duration * 60 * 60 * 1000 );
+      break;
+    case MDAL::Duration::days:
+      mDuration = int64_t( duration * 24 * 60 * 60 * 1000 );
+      break;
+    case MDAL::Duration::weeks:
+      mDuration = int64_t( duration * 7 * 24 * 60 * 60 * 1000 );
```
PeterPetrik: you shouldn't do this multiplication. Just multiply it on your calculator and create a static file variable like

const double SECONDS2HOURS = 1000;
const double DAYSS2HOURS = ?;

mdal/mdal_data_model.cpp
```diff
@@ -332,3 +337,293 @@ size_t MDAL::Mesh::faceVerticesMaximumCount() const
...
+    case MDAL::Duration::minutes:
+      mDuration = int64_t( duration * 60 * 1000 );
+      break;
+    case MDAL::Duration::hours:
+      mDuration = int64_t( duration * 60 * 60 * 1000 );
+      break;
+    case MDAL::Duration::days:
+      mDuration = int64_t( duration * 24 * 60 * 60 * 1000 );
+      break;
+    case MDAL::Duration::weeks:
```
PeterPetrik: do we need weeks?

mdal/mdal_data_model.cpp
```diff
@@ -332,3 +337,293 @@ size_t MDAL::Mesh::faceVerticesMaximumCount() const
...
+{
+  if ( !mValid || !other.mValid )
+    return Duration();
+  return Duration( mJulianTime - other.mJulianTime );
+}
+
+MDAL::Duration::Duration(): mDuration( 0 )
+{}
+
+MDAL::Duration::Duration( double duration, MDAL::Duration::Unit unit )
```
PeterPetrik: in general this is true, however UNIDATA (formats with CF/UGRID convention) are a bit tricky. see 

```udunits defines the periods as fixed multiples of seconds. The non-obvious ones are:

day = 86400.0 seconds
week = 7 days = 604800.0 seconds
month = year/12 = 2629743.831225 seconds
year = 3.15569259747E7 seconds
```

https://www.unidata.ucar.edu/software/netcdf-java/current/CDM/CalendarDateTime.html

you would need to have some special parameter in Duration DateTime to specify it is CF input and not normal one, so you can use different coefficient

mdal/mdal_data_model.cpp
```diff
@@ -332,3 +337,293 @@ size_t MDAL::Mesh::faceVerticesMaximumCount() const
...
+    msStr = std::string( "," ).append( msStr );
+  }
+
+  std::string strDateTime = prependZero( std::to_string( values.year ), 4 ) + "-" +
+                            prependZero( std::to_string( values.month ), 2 ) + "-" +
+                            prependZero( std::to_string( values.day ), 2 ) + "T" +
+                            prependZero( std::to_string( values.hours ), 2 ) + ":" +
+                            prependZero( std::to_string( values.minutes ), 2 ) + ":" +
+                            prependZero( std::to_string( int( values.seconds ) ), 2 ) +
+                            msStr; /// TODO use another way to translate seconds with fraction aprt
```
PeterPetrik: typo: part

mdal/frmts/mdal_binary_dat.cpp
```diff
@@ -78,6 +78,26 @@ static bool readIStat( std::ifstream &in, int sflg, char *flag )
   return false;
 }
 
+static MDAL::Duration convertTimeData( double time, const std::string &originalTimeDataUnit )
```
PeterPetrik: I am wondering if these conversion functions cannot be part of the new file with datatime in namespace like DataTimeUtils {} or so.. not sure if they are repeating or different for each format?

mdal/mdal_data_model.cpp
```diff
@@ -332,3 +337,293 @@ size_t MDAL::Mesh::faceVerticesMaximumCount() const
...
+{
+  if ( !mValid && !other.mValid )
+    return true;
+  return ( mValid && other.mValid ) && ( mJulianTime <= other.mJulianTime );
+}
+
+MDAL::DateTime::DateTime( int64_t julianTime ): mJulianTime( julianTime )
+{}
+
+MDAL::DateTime::DateTimeValues MDAL::DateTime::dateTimeGregorianJulianCalendar() const
```
PeterPetrik: in general, we would need to create some unit tests for this data time/ duration classes. I mean we need to test directly there classes alone, so we are sure the conversion between various time stuff is valid.

tests/mdal_testutils.cpp
```diff
@@ -312,3 +312,8 @@ void init_test()
 void finalize_test()
 {
 }
+
+bool compareDurationInHours( double h1, double h2 )
+{
+  return fabs( h1 - h2 ) < 1.0 / 3600 / 1000;
```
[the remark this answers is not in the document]
vcloarec: operator / is a left to right operator, so since the left operator is a float, the float division will be used.

mdal/mdal_data_model.hpp
vcloarec: I am not sure... My point of view :
Returning explicitly mean that the time is surely UTC. But, often, that's not the case and the time zone is unknown. 
When the time zone is not specified, that's mean the time is local (https://stackoverflow.com/questions/33114386/datetime-iso-8601-without-timezone-component).
In this implementation, time zone is not supported, so MDAL "doesn't know" the zone zone and return the time in local time.
If we decide that MDAL supports time zone (this is another question), if time zone is known, MDAL should return the time with time zone or UTC time, in conformity with ISO8601. If the tim zone is unknown, MDAL should return in local time.
Open for discussion...

mdal/mdal_data_model.hpp
vcloarec: Could be convenient int the future and used by `std::string toJulianDayString() const`

mdal/mdal_data_model.hpp
vcloarec: used in mdal_ascii_dat.cpp in `persist` method

mdal/mdal_data_model.hpp
vcloarec: My point of view:
Time can be confusing and can express a moment or a duration. 
For me QTime express a moment in the day :
"A QTime object contains a clock time, which it can express as the numbers of hours, minutes, seconds, and milliseconds since midnight." (https://doc.qt.io/qt-5/qtime.html#details)
Duration is more explicit and can't express a moment but only a duration.

mdal/mdal_data_model.cpp
vcloarec: Here, there are the time units which are fix in any context (as opposed to month and year). I don't know if we need now, but it can be used if needed...

mdal/mdal_data_model.cpp
vcloarec: I am not sure to understand... 
The CF/UGRID convention is consistent with day=86400 seconds, and weeks=7 days, so why treat this unit differently.
For year/month, we can adopt the CF convention or not. Right now, the user has to choose how to convert year/month unit by converting himself before calling this constructor.

mdal/frmts/mdal_binary_dat.cpp
vcloarec: I agree if formats treat units is same way, but if the don't, I think particular treatment have to be in the file of the format.

mdal/mdal_data_model.cpp
vcloarec: I agree, tests will be an important part of the work!

tests/mdal_testutils.cpp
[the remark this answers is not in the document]
PeterPetrik: I understand, but better to evaluate all beforehand to one single constant and use that one. also adding .0 helps readability

mdal/mdal_data_model.hpp
PeterPetrik: i would keep it as simple as possible

mdal/mdal_data_model.hpp
PeterPetrik: in general, we implement things when we need them

mdal/mdal_data_model.hpp
PeterPetrik: ah ok

mdal/mdal_data_model.hpp
PeterPetrik: this is naming, i leave it with you

mdal/mdal_data_model.cpp
PeterPetrik: no problem, just to keep it in mind when working with formats based on CF.

vcloarec:
@wonder-sk I will be happy if you have linked toward this article

mdal/mdal_data_model.cpp
vcloarec: Finally, I have implemented year and month too. This is necessary if CF driver meet this time unit. Duration class could adopts UNIDATA. Not sure... I have to think about...

mdal/frmts/mdal_binary_dat.cpp
vcloarec: Finally, I have implemented a unique function in MDAL::utils that parses the unit string to retrieve the unit.

mdal/frmts/mdal_cf.cpp
```diff
@@ -161,7 +161,78 @@ static void populate_vals( bool is_vector, double *vals, size_t i,
   }
 }
 
-void MDAL::DriverCF::addDatasetGroups( MDAL::Mesh *mesh, const std::vector<double> &times, const MDAL::cfdataset_info_map &dsinfo_map )
+static MDAL::DateTime parseReferenceTime( std::string timeInformation, std::string calendar )
```
PeterPetrik: const std::string&

wonder-sk:
@vcloarec unfortunately I can't find the original article... but this one also give quite a good list about some false assumptions programmers make about time: https://infiniteundo.com/post/25326999628/falsehoods-programmers-believe-about-time

tests/test_sww.cpp
```diff
@@ -151,6 +151,13 @@ TEST( MeshSWWTest, Cairns )
   EXPECT_DOUBLE_EQ( 0, min );
   EXPECT_DOUBLE_EQ( 6.7305092811584473, max );
 
+  const char *referenceTime;
+  referenceTime = MDAL_G_referenceTime( g );
+  EXPECT_EQ( std::string( "none" ), std::string( referenceTime ) );
```
PeterPetrik: i would return empty string in case format/dataset does not have reference time at all

tests/test_xdmf.cpp
```diff
@@ -112,7 +116,11 @@ TEST( XdmfTest, Basement3HumpsTest )
     EXPECT_DOUBLE_EQ( 4.9994493491047303, max );
 
     double time = MDAL_D_time( ds );
-    EXPECT_DOUBLE_EQ( 20, time );
+    EXPECT_TRUE( compareDurationInHours( 20, time ) );
+
+    const char *referenceTime;
```
PeterPetrik: I am thinking if you can create a function in mdal test utils hasReferenceTime( group) that returns true if reference time != 0

and in all tests just add 

EXPECT_FALSE( hasReferenceTime(g) ); 

instead of these 3-4 lines

mdal/frmts/mdal_xdmf.cpp
```diff
@@ -435,8 +434,7 @@ MDAL::DatasetGroups MDAL::DriverXdmf::parseXdmfXml( )
   {
     ++nTimesteps;
     xmlNodePtr timeNod = xmfFile.getCheckChild( gridNod, "Time" );
-    double time = xmfFile.queryDoubleAttribute( timeNod, "Value" );
-
+    Duration time( xmfFile.queryDoubleAttribute( timeNod, "Value" ), Duration::hours ); //units, hours ?
```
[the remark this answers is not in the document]
vcloarec: xdmf specification do not required time unit and in the test file, there is no time unit specified. if any doubt, what to do ? Which model use the format, only Basement ? Does this model use always hours ? Same issue than ASCII_DAT.
Use a unknown time unit flag ?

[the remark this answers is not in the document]
PeterPetrik: do it that way that it is not creating any regression. So it will give same results as previous MDAL versions.

mdal/mdal_date_time.hpp
```diff
@@ -0,0 +1,143 @@
...
+      bool operator<( const Duration &other ) const;
+      bool operator>( const Duration &other ) const;
+      bool operator>=( const Duration &other ) const;
+      bool operator<=( const Duration &other ) const;
+
+    private:
+      Duration( int64_t ms );
+      int64_t mDuration; //in ms
+
+      friend class DateTime;
```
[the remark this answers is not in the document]
vcloarec: Not sure... those classes are "real friends" and I don't want to expose `int64_t mDuration` used for operation between `DateTime` and `Duration`. `int64_t mDuration` is now used to store the information but maybe, one day, it will not be anymore (for example, if we used a third library). So I don't want to expose this value.

mdal/mdal_utils.cpp
```diff
@@ -637,3 +637,138 @@ std::string MDAL::doubleToString( double value, int precision )
...
+    return  str;
+
+  return std::string( length - str.size(), '0' ).append( str );
+}
+
+MDAL::Duration::Unit MDAL::parseDurationTimeUnit( const std::string &timeUnit )
+{
+  MDAL::Duration::Unit unit = MDAL::Duration::hours; //default unit
+
+  if ( timeUnit == "millisec" ||
```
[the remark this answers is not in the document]
vcloarec: for plural form or abbreviation ?

tests/test_ascii_dat.cpp
```diff
@@ -191,8 +191,9 @@ TEST( MeshAsciiDatTest, QuadAndTriangleFaceScalarFile )
   double time = MDAL_D_time( ds );
   EXPECT_DOUBLE_EQ( 1, time );
 
-  const char *referenceTime = MDAL_G_referenceTime( g );
-  EXPECT_EQ( std::string( "JULIAN 2433282.500000" ), std::string( referenceTime ) );
+  const char *referenceTime;
+  referenceTime = MDAL_G_referenceTime( g );
+  EXPECT_EQ( std::string( "1950-01-01T00:00:00" ), std::string( referenceTime ) );
```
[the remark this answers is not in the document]
vcloarec: from online converter. But good idea to generate julian day with Qt to add tests

tests/test_hec2d.cpp
```diff
@@ -144,6 +144,11 @@ TEST( MeshHec2dTest, simpleArea )
   double time = MDAL_D_time( ds );
   EXPECT_DOUBLE_EQ( 0.0, time );
 
+  EXPECT_TRUE( compareDurationInHours( 0, time ) );
+
+  const char *referenceTime = MDAL_G_referenceTime( g );
+  EXPECT_EQ( std::string( "1899-12-30T00:00:00" ), std::string( referenceTime ) );
```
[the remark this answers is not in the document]
vcloarec: ?

mdal/frmts/mdal_gdal_grib.cpp
```diff
@@ -39,21 +39,23 @@ bool MDAL::DriverGdalGrib::parseBandInfo( const MDAL::GdalDataset *cfGDALDataset
...
+    mRefTime = DateTime( parseMetadataTime( iter->second ), DateTime::Unix );
   }
 
   // TIME
   iter = metadata.find( "grib_valid_time" );
   if ( iter == metadata.end() ) return true; //FAILURE
-  double valid_time = parseMetadataTime( iter->second );
-  *time = ( valid_time - mRefTime ) / 3600.0; // input times are always in seconds UTC, we need them back in hours
+  DateTime valid_time = DateTime( parseMetadataTime( iter->second ), DateTime::Unix );
+  *time = ( valid_time - mRefTime );
```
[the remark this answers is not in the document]
vcloarec: tested indirectly here :
https://github.com/lutraconsulting/MDAL/blob/c948bdac0511792535629e8a26e9f722d26f23ca/tests/test_gdal_grib.cpp#L59

mdal/mdal_data_model.hpp
```diff
@@ -75,14 +76,15 @@ namespace MDAL
       DatasetGroup *group() const;
       Mesh *mesh() const;
 
-      double time() const;
-      void setTime( double time );
+      double time( Duration::Unit unit = Duration::hours ) const;
+      void setTime( double time, Duration::Unit unit = Duration::hours );
```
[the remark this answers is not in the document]
vcloarec: I had this default to be backward compatible with old code during work. Now, no need anymore

PR #180, opened 2019-12-09
Active flag refactoring

PeterPetrik:
- adds support for tuflowfv active flags
- adds support for "active flag capability"

### DIFF
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -292,13 +292,15 @@ MDAL_EXPORT void MDAL_G_minimumMaximum( DatasetGroupH group, double *min, double
 //!
 //! Minimum and maximum dataset values are automatically calculated
 //!
+//! Only for 2D datasets
+//!
 //! \param group parent group handle
 //! \param time time for dataset
 //! \param values For scalar data on vertices, the size must be vertex count
 //!               For scalar data on faces, the size must be faces count
 //!               For vector data on vertices, the size must be vertex count * 2 (x1, y1, x2, y2, ..., xN, yN)
 //!               For vector data on faces, the size must be faces count * 2 (x1, y1, x2, y2, ..., xN, yN)
-//! \param active if null pointer, all faces are active. Otherwise size must be equal to face count.
+//! \param active if null pointer, MDAL_D_hasActiveFlagCapability returns false. Otherwise size must be equal to face count.
 //! \returns empty pointer if not possible to create dataset (e.g. group opened in read mode), otherwise handle to new dataset
 MDAL_EXPORT DatasetH MDAL_G_addDataset( DatasetGroupH group,
                                         double time,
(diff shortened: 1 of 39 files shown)


mdal/frmts/mdal_flo2d.cpp
```diff
@@ -253,17 +253,15 @@ void MDAL::DriverFlo2D::parseTIMDEPFile( const std::string &datFileName, const s
       if ( face_idx == nVertexs ) throw MDAL_Status::Err_IncompatibleMesh;
 
       // this is magnitude: getDouble(lineParts[2]);
-      flowDataset->values()[2 * face_idx] = getDouble( lineParts[3] );
-      flowDataset->values()[2 * face_idx + 1] = getDouble( lineParts[4] );
+      flowDataset->setValue( face_idx, getDouble( lineParts[3] ),  getDouble( lineParts[3] ) );
```
wonder-sk: the second value should be lineParts[4] ?

mdal/mdal_memory_data_model.hpp
```diff
@@ -34,29 +36,47 @@ namespace MDAL
...
+       *
+       * Dataset must support active flags
+       */
+      void setActive( size_t index, int stat );
+      void setActive( const int *activeBuffer );
+      int active( size_t index ) const;
+
+
+      void setValue( size_t index, double value );
+      void setValue( size_t index, double x, double y );
```
wonder-sk: It is a bit counter-intuitive that these setValue() calls work in a very different way - one uses index, while the other one uses 2*index and 2*index+1 ... how about making it more explicit - e.g. setScalarValue( index, value ) and setVectorValue( index, x, y )  ? There could be also setVector[X|Y]Value( index, value ) for the cases where we set X/Y value separately? Because for example in Selafin driver it looked like a bug that we call setValue(2*i, ...) when we want to set X/Y value for index i.

wonder-sk: It would be good to put implementation of these methods to the header file so they can be inlined by the compiler... (the bodies are very simple anyway)

mdal/api/mdal.h
```diff
@@ -347,18 +349,20 @@ MDAL_EXPORT int MDAL_D_valueCount( DatasetH dataset );
 //! Returns whether dataset is valid
 MDAL_EXPORT bool MDAL_D_isValid( DatasetH dataset );
 
+//! Returns whether dataset supports active flag for dataset faces
+MDAL_EXPORT bool MDAL_D_hasActiveFlagCapability( DatasetH dataset );
```
wonder-sk: Maybe it would be better to use a mix of capability flags... e.g. `int MDAL_D_capabilities(dataset)` - and each capability would be a different bit? We would not need to add extra exported methods to the API...

But I am easy - we can also leave it as is...

mdal/mdal_memory_data_model.hpp
```diff
@@ -34,29 +36,47 @@ namespace MDAL
...
+      double value( size_t index ) const;
```
wonder-sk: Also for the getter - it would be good to have scalar/vector specific getters.

Also, the getters could check for the dataset type and use assert to make sure we don't call vectorXValue(index) on a scalar dataset and vice versa...

mdal/api/mdal.h
PeterPetrik: we have this version, also for drivers etc, see `MDAL_DR_saveMeshCapability` . lets keep it this way (Consistent) we can always change in the future

PR #186, opened 2019-12-16
Introduce unknown time unit

vcloarec:
New approach when time unit is unknown, for discussion :
Where there is a doubt for the time unit, MDAL considers the time unit as unknown and the time unit is stored for all format in the metadata of the dataset group.
So before asking for the time, QGIS can check if the time unit is known or not. If not, he will ask the user to choose the time unit or a factor to convert the time in hours, as QGIS read it.


PeterPetrik:
I would suggest to do not add this change to the API for the next release. I will solve the timeunits problem on QGIS side.

PeterPetrik:
I am closing this for now, we can reopen it and rebase anytime later if we decide to add this option to MDAL (I am free to discussion on chat)

PR #234, opened 2020-03-23
Create MDAL_MeshNames API method

tomasMizera:
PR adds method `const char *MDAL_MeshNames( const char *filePath );` inside `mdal.h`  API

All drivers besides UGRID return string in format:  `<drivername>:"<meshfile>"` if file exists but UGRID driver returns string in format `Ugrid:"<meshfile>":<meshname>`. If more meshes are included in file, separator `;;` is used and the same string is repeated with different meshname, such as:
`Ugrid:"<meshfile>":<meshname1>;;Ugrid:"<meshfile>":<meshname2>`


.gitignore
```diff
@@ -37,3 +37,4 @@ qgis-test.ctest
 qtcreator-build/
 scripts/Debug
 Thumb.db
+cmake-build-debug/*
```
tomasMizera: This is for my own builds, may I commit it or not?

PeterPetrik: you can keep it in gitignore, np

mdal/mdal.cpp
```diff
@@ -174,6 +174,14 @@ MDAL_MeshH MDAL_LoadMesh( const char *uri )
     return static_cast< MDAL_MeshH >( MDAL::DriverManager::instance().load( meshFile, meshName ).release() );
 }
 
+const char *MDAL_MeshNames( const char *filePath )
+{
+  std::string file( filePath );
```
PeterPetrik: what happens if filePath is nullptr?

mdal/mdal_driver_manager.cpp
```diff
@@ -38,6 +38,32 @@
...
+  }
+
+  for ( const auto &driver : mDrivers )
+  {
+    if ( ( driver->hasCapability( Capability::ReadMesh ) ) &&
+         driver->canReadMesh( file ) )
+    {
+      std::unique_ptr<Driver> drv( driver->create() );
+
+      if ( drv->load( file ) )
```
PeterPetrik: we do not want to use load(file) here. canReadMesh() is enough... 

`canReadMesh` is fast function that should tell you exactly that file can be loaded. 
`load` is very slow function (sometimes) and you do not want to read 10GB of text file to throw it away second later :)

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -45,6 +45,27 @@ std::vector<std::string> MDAL::DriverUgrid::findMeshesNames() const
...
+  std::string meshUris( "" );
+
+  for ( size_t it = 0; it < meshNames.size(); ++it )
+  {
+    meshUris += name() + ":\"" + mFileName + "\":" + meshNames.at( it );
+    if ( ( it + 1 ) < meshNames.size() ) // If this is not the last mesh in array, add separator
+      meshUris += MDAL_URI_SEPARATOR;
+  }
+
+  return meshUris;
```
PeterPetrik: I think this can go to mdal_utils.hpp where as arguments you will pass vector or uris and it will merge them.
also maybe create function in mdal utils that take driver name, file name and optionally meshName to create uri?

tests/test_api.cpp
```diff
@@ -284,6 +284,42 @@ TEST( ApitTest, LoggerApi )
...
 }
 
+TEST( ApitTest, BuildUrisApi )
+{
+  MDAL_SetLoggerCallback( &_testLoggerCallback );
+  MDAL_SetLogVerbosity( MDAL_LogLevel::Debug );
+
+  std::vector<std::pair<std::string, std::string>> testScenarios
+  {
+    {
```
PeterPetrik: can you add that 1d 2d ugrid file please?

tests/test_api.cpp
```diff
@@ -284,6 +284,42 @@ TEST( ApitTest, LoggerApi )
...
+      test_file( "/2dm/regular_grid.2dm" ),
+      "2DM:\"" + test_file( "/2dm/regular_grid.2dm" ) + "\""
+    },
+    {
+      test_file( "/ugrid/1dtest/dflow1d_map.nc" ),
+      "Ugrid:\"" + test_file( "/ugrid/1dtest/dflow1d_map.nc" ) + "\":" + "network" + ";;"
```
PeterPetrik: I am still thinking if we shouldn't exclude "network". it looks like it is never real mesh so I would just not list it from UGRID driver at all. if someone complains, we can add it back...

PeterPetrik:
1. meshNames can take DRIVER:path to force only one driver
2. function in test_utils to test meshNames ---> we test that canLoad pick the correct driver.

mdal/mdal.cpp
tomasMizera: :bug: :bug:

tests/test_api.cpp
```diff
@@ -284,6 +284,52 @@ TEST( ApitTest, LoggerApi )
   EXPECT_EQ( receivedLogMessage, "No driver with index: -1" );
 }
 
+TEST( ApiTest, BuildUrisApi )
```
PeterPetrik: this is tested in the driver tests directly, not sure if we need it here at all. in the test_api.cpp you should test more some iff cases like invalid filename, empty filename, non-existent driver name and so on...

mdal/mdal_driver_manager.cpp
```diff
@@ -80,15 +122,18 @@ std::unique_ptr<MDAL::Mesh> MDAL::DriverManager::load(
...
+  requestedDriver = driver( driverName );
+  if ( !requestedDriver )
   {
     MDAL::Log::error( MDAL_Status::Err_MissingDriver, "Could not find driver with name: " + driverName );
     return mesh;
   }
 
-  mesh = dr->load( meshFile, meshName );
+  std::unique_ptr<Driver> drv( requestedDriver->create() );
+  if ( drv->hasCapability( Capability::ReadMesh ) && drv->canReadMesh( meshFile ) )
```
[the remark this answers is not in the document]
tomasMizera: But what if user makes a mistake? (by choosing wrong driver)

mdal/mdal_utils.cpp
```diff
@@ -854,3 +873,42 @@ void MDAL::parseDriverAndMeshFromUri( const std::string &uri, std::string &drive
   parseMeshFileFromUri( uri, meshFile );
   parseSpecificMeshFromUri( uri, meshName );
 }
+
+std::string MDAL::buildMeshUri( const std::string &meshFile, const std::string &meshName, const std::string &driver )
```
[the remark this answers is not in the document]
tomasMizera: Well, the meshfile is trimmed from quotes in MDAL_MeshNames calling the function `MDAL::parseMeshFileFromUri( uriString, file );`. However, it is true that we do not need to add quotes

mdal/mdal_utils.cpp
```diff
@@ -854,3 +873,42 @@ void MDAL::parseDriverAndMeshFromUri( const std::string &uri, std::string &drive
...
+{
+  mergedUris.clear();
+  size_t meshNamesCount = meshNames.size();
+
+  for ( size_t it = 0; it < meshNamesCount; ++it )
+  {
+    mergedUris += buildMeshUri( meshFile, meshNames.at( it ), driver );
+
+    if ( ( it + 1 ) < meshNamesCount ) // If this is not the last mesh in array, add separator
+      mergedUris += MDAL_URI_SEPARATOR;
```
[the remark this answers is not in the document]
tomasMizera: It is, I was just thinking about not having random constants in code.

mdal/mdal_utils.hpp
```diff
@@ -36,7 +38,8 @@ namespace MDAL
 
   /** Return whether file exists */
   bool fileExists( const std::string &filename );
-  std::string baseName( const std::string &filename );
+  std::string baseName( const std::string &filename, const bool &keepExtension = false );
```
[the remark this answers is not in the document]
tomasMizera: Very true, I guess it was too late :D

tests/test_api.cpp
tomasMizera: I added more edge cases

PR #250, opened 2020-05-05
classification option for UGRID

vcloarec:


mdal/frmts/mdal_3di.cpp
```diff
@@ -194,7 +194,13 @@ std::set<std::string> MDAL::Driver3Di::ignoreNetCDFVariables()
...
 }
 
-void MDAL::Driver3Di::parseNetCDFVariableMetadata( int varid, const std::string &variableName, std::string &name, bool *is_vector, bool *is_x )
+void MDAL::Driver3Di::parseNetCDFVariableMetadata(
+  int varid,
+  const std::string &variableName,
+  std::string &name,
+  bool *is_vector,
+  bool *is_x,
+  std::map<std::string, std::string> & )
```
PeterPetrik, line 203: add options explicitely and use MDAL_UNUSED in the function body https://github.com/lutraconsulting/MDAL/blob/417898347448e1dab7decc2d31a9c418873f9513/mdal/mdal_utils.hpp#L23

mdal/frmts/mdal_3di.hpp
```diff
@@ -51,7 +51,8 @@ namespace MDAL
       std::string getTimeVariableName() const override;
       std::set<std::string> ignoreNetCDFVariables() override;
       void parseNetCDFVariableMetadata( int varid, const std::string &variableName,
-                                        std::string &name, bool *is_vector, bool *is_x ) override;
+                                        std::string &name, bool *is_vector, bool *is_x,
+                                        std::map<std::string, std::string> &option ) override;
```
PeterPetrik, line 55: option -> options

PeterPetrik: Wouldn't be possible to use directly `Metadata` class https://github.com/lutraconsulting/MDAL/blob/417898347448e1dab7decc2d31a9c418873f9513/mdal/mdal_data_model.hpp#L40

mdal/frmts/mdal_cf.cpp
```diff
@@ -176,6 +177,9 @@ void MDAL::DriverCF::addDatasetGroups( MDAL::Mesh *mesh, const std::vector<Relat
         );
     group->setIsScalar( !dsi.is_vector );
 
+    for ( auto const &elem : dsi.options )
+      group->setMetadata( elem.first, elem.second );
+
```
PeterPetrik, line 182: I would just add new function `void setMetadata(const Metadata& meta)`
https://github.com/lutraconsulting/MDAL/blob/417898347448e1dab7decc2d31a9c418873f9513/mdal/mdal_data_model.hpp#L152

mdal/frmts/mdal_tuflowfv.cpp
```diff
@@ -456,7 +456,12 @@ std::set<std::string> MDAL::DriverTuflowFV::ignoreNetCDFVariables()
   return ignore_variables;
 }
 
-void MDAL::DriverTuflowFV::parseNetCDFVariableMetadata( int varid, const std::string &variableName, std::string &name, bool *is_vector, bool *is_x )
+void MDAL::DriverTuflowFV::parseNetCDFVariableMetadata(
+  int varid,
+  const std::string &variableName,
+  std::string &name,
+  bool *is_vector,
+  bool *is_x, std::map<std::string, std::string> & )
```
PeterPetrik, line 464: each arg on separate line, use MDAL_UNUSED

mdal/frmts/mdal_tuflowfv.hpp
```diff
@@ -123,7 +123,8 @@ namespace MDAL
       std::string getCoordinateSystemVariableName() override;
       std::set<std::string> ignoreNetCDFVariables() override;
       void parseNetCDFVariableMetadata( int varid, const std::string &variableName,
-                                        std::string &name, bool *is_vector, bool *is_x ) override;
+                                        std::string &name, bool *is_vector, bool *is_x,
+                                        std::map<std::string, std::string> &option ) override;
```
PeterPetrik, line 127: options

tests/data/ugrid/classified/README.txt
```diff
@@ -0,0 +1,19 @@
...
+*Domain:* 2D Cartesian grid, 100x40m extent. 10x4 square grid cells.
+Coordinate reference system is not relevant here. 
+
+*Data:*
+* "mesh2d_s1": Water level in cell centres, as classes.
+* "mesh2d_waterdepth": Water depth in cell centres, as classes.
+
+*Time:*   24 snapshots in time.
+
+NOTE: replaced on 6th April 2020, now it has new “:flag_bounds” attribute that specifies the connection
```
PeterPetrik, line 19: remove this NOTE LINE

tests/test_ugrid.cpp
```diff
@@ -619,6 +619,26 @@ TEST( MeshUgridTest, 1DMeshTest )
   MDAL_CloseMesh( m );
 }
 
+TEST( MeshUgridTest, classifiedVariable )
+{
+  std::string path = test_file( "/ugrid/classified/simplebox_clm.nc" );
+  MDAL_SetLogVerbosity( MDAL_LogLevel::Debug );
```
PeterPetrik, line 625: Log verbosity should not be set by individual tests

tests/test_ugrid.cpp
```diff
@@ -619,6 +619,26 @@ TEST( MeshUgridTest, 1DMeshTest )
...
+  MDAL_MeshH mesh = MDAL_LoadMesh( path.c_str() );
+
+  EXPECT_FALSE( mesh == nullptr );
```
PeterPetrik, line 628: EXPECT_NE( m, nullptr );

PeterPetrik:
1. can you please bump the version of MDAL to 0.5.92?
2. I still think we need to parse `flag_meanings`, see https://content.oss.deltares.nl/delft3d/manuals/D-Flow_FM_User_Manual.pdf#subsubsection.B.2.11.9

> 
> This section is not specific for 1D network input, but applies to describing packed data on any
> mesh or network. An example use of this is the classmap output as described in Section F.3.3.
> Output data may be compressed by packing, which is explained in the CF-conventions as
> "altering the data in a way that reduces its precision" 4http://cfconventions.org/cf-conventions/cf-conventions.html#packed-data
> . For example, floating point data may be compressed by categorizing it in consecutive intervals, each represented by an integer class number. All classes together span the full data range.
> Here we propose a small extension to the CF-conventions concept of flag values and flag meanings. We introduce a new data variable attribute flag_bounds, which gives the name of a variable containing the bounds of each category. This flag bounds variable for a set of N contiguous intervals must be an array of shape (N,2), where N is the slowest varying dimension. When an interval has an open end, that bound should be set to the fill value. The following example packs a velocity magnitude variable into 8 intervals of non-uniform size, the last of which open-ended.
> dimensions:
> Two = 2 ;
> mesh2d_nFaces = 40 ;
> time = UNLIMITED ; // (24 currently) class_ucmag = 8 ;
> variables:
> byte mesh2d_ucmag(time, mesh2d_nFaces) ;
> mesh2d_ucmag:mesh = "mesh2d" ;
> mesh2d_ucmag:location = "face" ;
> mesh2d_ucmag:coordinates = "mesh2d_face_x mesh2d_face_y" ; 
> mesh2d_ucmag:cell_methods = "mesh2d_nFaces: mean" ; 
> mesh2d_ucmag:cell_measures = "area: mesh2d_flowelem_ba" ; 
> mesh2d_ucmag:standard_name = "sea_water_speed" ; 
> mesh2d_ucmag:long_name = "Flow element center velocity magnitude" ; 
> mesh2d_ucmag:units = "m s-1" ;
> mesh2d_ucmag:grid_mapping = "projected_coordinate_system" ;
> mesh2d_ucmag:flag_values = 1b, 2b, 3b, 4b, 5b, 6b, 7b, 8b ; 
**> mesh2d_ucmag:flag_bounds = "class_bounds_ucmag" ; mesh2d_ucmag:flag_meanings = "0.000m_s-1_to_0.200m_s-1
> 0.200m_s-1_to_0.400m_s-1 0.400m_s-1_to_0.600m_s-1 0.600m_s-1_to_0.800m_s-1 0.800m_s-1_to_1.000m_s-1 1.000m_s-1_to_1.500m_s-1 1.500m_s-1_to_2.000m_s-1 above_2.000m_s-1" ;**
> double class_bounds_ucmag(class_ucmag, Two) ;
> data: mesh2d_ucmag =
>     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
>       1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1,
>     1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 1, 1,
>       1, 4, 2, 1, 1, 5, 4, 2, 1, 5, 4, 2, 5, 4, 5,
> //... class_bounds_ucmag =
> 0, 0.2, 0.2, 0.4, 0.4, 0.6, 0.6, 0.8, 0.8, 1, 1, 1.5, 1.5, 2, 2, _ ;

vcloarec:
>     1. can you please bump the version of MDAL to 0.5.92?

Maybe we can waiting to implement reading mag/direction before bump the version?



>     2\. I still think we need to parse `flag_meanings`

We have also : 
```
class_bounds_ucmag =
0, 0.2, 0.2, 0.4, 0.4, 0.6, 0.6, 0.8, 0.8, 1, 1, 1.5, 1.5, 2, 2, _ ;
```
and here : https://gitlab.cloud.lutraconsulting.co.uk/lutraconsulting/mdal-kanban/issues/18#note_16732.
There is:

> The labels here are ok. But with the addition of the newly introduced :flag_bounds, you could also consider pretty printing the label for each category using the upper and lower bound floating point value, such that you don’t have the (mandatory) underscores in the labels as they come from the :flag_meanings attribute.
 


>     3\. I think we do not need to parse units, it is part of the flag_meanings

I find convenient to get the unit here. Futhermore, it is parse not only for classified value, so it could be used too for other usage. In QGIS, it could be used with the color ramp shader  with default style to put the unit in the label.

>     4\. I would the metadata variable name should be more unique (so it does not clash in future with some random metadata key), e.g. `mdal_classification` (or similar)

ok with that



>     5\. also it is good to check  QgsColorRampShader::ColorRampItem if there is no string format already defined for it?

In QGIS, the label formating is only used when loading the mesh for the first time (default settings). If the color ramp shader is reclassified (so the unit label defined in the color ramp shader is used), the classified labeling is not valid anymore.

PeterPetrik:
ok, if the labels can generated from class_bounds_ucmag + units it is OK as current implementation of those

PR #263, opened 2020-06-01
detect from/to direction for polar vector

### LINKED ISSUE #254: [UGRID] Detect polar vectors from standard name
similar to long_name, use substring `_from/to_direction`

https://github.com/lutraconsulting/MDAL/blob/9c5b479f8915bf0ff239276a220196118d306771/mdal/frmts/mdal_ugrid.cpp#L550

vcloarec:
waiting good test files
fix #254


mdal/frmts/mdal_ugrid.cpp
```diff
@@ -525,6 +527,27 @@ void MDAL::DriverUgrid::parseNetCDFVariableMetadata( int varid,
         *isX = false;
         name = MDAL::replace( standardName, "_y_", "" );
       }
+      else if ( MDAL::contains( standardName, "_speed" ) )
+      {
+        *isVector = true;
+        *isPolar = true;
+        name = MDAL::replace( standardName, "_speed", "_velocity" );
```
PeterPetrik, line 534: Why we want to replace speed with velocity? i would keep it as in original data

tests/test_ugrid.cpp
```diff
@@ -665,7 +665,7 @@ TEST( MeshUgridTest, magnitude_direction )
   int magIndex = getValue( dataset, 20 );
   EXPECT_EQ( magIndex, 3 );
 
-  double direction = ( directionIndex - 1 ) * 10 + 5;
+  double direction = -( ( directionIndex - 1 ) * 10 + 5 ); //supposed to be clokcwise angle
```
PeterPetrik, line 668: clockwise

mdal/frmts/mdal_cf.hpp
```diff
@@ -60,8 +60,9 @@ namespace MDAL
     };
     std::string name; //!< Dataset group name
     CFDimensions::Type outputType;
-    bool is_vector;
-    bool is_polar;
+    bool isVector;
+    bool isPolar;
+    bool isInvertedDirection;
```
PeterPetrik, line 65: maybe rename to `hasInvertedPolarDirection`? (i assume it is applicable only for polar dir?)

mdal/frmts/mdal_ugrid.cpp
vcloarec: Because the direction name contain velocity and, given the choice, it is preferable to be consistent with other possible variable name.

PeterPetrik: I would keep their names. It is their format and their name choices for any reasons they have...

vcloarec: ok. So with direction : velocity_from_direction --> speed

PeterPetrik:
why was this closed and not merged?

PeterPetrik:
i cannot reopen the PR too since the original branch is deleted

Issue #259, opened 2020-05-22
Material ID support for 2DM meshes


PeterPetrik:
see https://github.com/qgis/QGIS/issues/36579

PR #264, opened 2020-06-02
Add MATID support for 2DM meshes

leonhard-s:
This implements some of the changes I mentioned in #259 / qgis/QGIS#36579.

I extended the system that adds the optional, secondary MATID column used by BASEMESH 3 as a dataset group to also support the primary MATID, as well as any number of additional MATIDs as specified by the `NUM_MATERIALS_PER_ELEM` card of the 2DM file.

The first Material ID will be added under the name `MATID`, the second keeps the the name `Bed Elevation (Face)` from the previous implementation. Any MATIDs beyond that receive the generic name `Auxiliary Material ID (counter)`.

This had two side effects that required modifications to the corresponding tests:

- The `NUM_MATERIALS_PER_ELEM` card is no longer optional. It used to be ignored entirely, now it must be provided for any MATIDs to be processed.

- The order of the created dataset groups has changed slightly.
    It used to be `Bed Elevation (Face)` (if available), then `Bed Elevation`, then any dataset groups added by external sources.
    In the new implementation, it's `Bed Elevation` first, then `Material ID`, then `Bed Elevation (Face)` (since it takes the place of the second MATID), then any other MATIDs, and finally the external dataset groups.


tests/data/xdmf/basement3/3HumpsTest/3_humps_mesh.2dm
```diff
@@ -1,4 +1,5 @@
 MESH2D   #created automatically via meshModel tool
+NUM_MATERIALS_PER_ELEM 2
```
PeterPetrik, line 2: The test source files cannot be unfortunately modified. The test files comes from the runs of the solvers and they are not usually made by hand. So modifying them means that the we are not testing the real world example.

mdal/frmts/mdal_2dm.cpp
```diff
@@ -187,26 +193,30 @@ std::unique_ptr<MDAL::Mesh> MDAL::Driver2dm::load( const std::string &meshFile,
...
 
       for ( size_t i = 0; i < faceVertexCount; ++i )
         face[i] = MDAL::toSizeT( chunks[i + 2] ) - 1; // 2dm is numbered from 1
 
-      // OK, now find out if there is optional cell elevation (BASEMENT 3.x)
-      if ( chunks.size() == faceVertexCount + 4 )
-      {
+      // This assertion will fail if a mesh has fewer material ID columns than
+      // promised by the NUM_MATERIAL_PER_ELEM tag.
+      assert( chunks.size() - 5 >= materialCount );
```
PeterPetrik, line 206: To support BASEMENT 3.x, I would do a logic a bit different here. 

`if chunks.size() - 5 = materialCount` -> current implementaion
`if chunks.size() - 5 = materialCount` -> take the one "extra" columen as `elevation - optional` (even NUM_MATERIALS_PER_ELEM==0)

tests/data/xdmf/basement3/3HumpsTest/3_humps_mesh.2dm
leonhard-s: Ah, of course - I'll revert them to their previous versions and add support for face elevation without having the tag set, then.

mdal/frmts/mdal_2dm.cpp
```diff
@@ -187,26 +195,46 @@ std::unique_ptr<MDAL::Mesh> MDAL::Driver2dm::load( const std::string &meshFile,
...
       // Right now we just store node IDs here - we will convert them to node indices afterwards
       assert( chunks.size() > faceVertexCount + 1 );
 
       for ( size_t i = 0; i < faceVertexCount; ++i )
         face[i] = MDAL::toSizeT( chunks[i + 2] ) - 1; // 2dm is numbered from 1
 
-      // OK, now find out if there is optional cell elevation (BASEMENT 3.x)
-      if ( chunks.size() == faceVertexCount + 4 )
+      // NUM_MATERIALS_PER_ELEM tag provided, use new MATID parser
+      if ( newMaterialParser )
```
PeterPetrik, line 207: if (materialCount > =0) && (chunks.size() - 5 >= materialCount) ...

mdal/frmts/mdal_2dm.cpp
```diff
@@ -125,6 +125,8 @@ std::unique_ptr<MDAL::Mesh> MDAL::Driver2dm::load( const std::string &meshFile,
   size_t faceCount = 0;
   size_t vertexCount = 0;
   size_t edgesCount = 0;
+  size_t materialCount = 0;
+  bool newMaterialParser = false;
```
PeterPetrik, line 129: please rename to `hasMaterialsDefinitionsForElements`

PR #266, opened 2020-06-13
[WIP] support writting with active flag

vcloarec:
Working on saving dataset on QGIS side, it appeared that it was no possible to write datasets on faces when those datasets has active flag. 
Maybe I missed something, but I do not understand why? Maybe an old logic that is not valid anymore?
For example TUFLOWFV read data on face with active flag.
This PR allows writing dataset on faces with active flag.

### DIFF
--- a/mdal/frmts/mdal_binary_dat.cpp
+++ b/mdal/frmts/mdal_binary_dat.cpp
@@ -464,7 +464,7 @@ bool MDAL::DriverBinaryDat::persist( MDAL::DatasetGroup *group )
       // Write status flags
       for ( size_t i = 0; i < elemCount; i++ )
       {
-        bool active = static_cast<bool>( dataset->active( i ) );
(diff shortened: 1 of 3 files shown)


PeterPetrik:
It was always the case that the active flag concept was present only for data defined on vertices. 
the need is because even you do have defined all the values on vertices, you may want to manually switch off the face (e.g. there is a house on that face area, so you want to omit it alltogether, but you need values defined on vertices so the adjacent faces can use them). 

for data defined on faces, you can use _FillValue/NODATA concept, so the active flag is not needed. 

TUFLOW-FV are usually 3D volume data, so maybe you try to save the 3D data? 
I think QGIS should never try to persist the active flag for data defined on faces,.... If it does, it is a bug in qgis I suppose

PeterPetrik:
this has WIP prefix

vcloarec:
for now, not sure to keep it...

PeterPetrik:
> for now, not sure to keep it...

?

PR #267, opened 2020-06-13
Fix compiler warnings about unused variables (on QGIS Docker builds)

rouault:


PeterPetrik:
as it is already mergin in QGIS anyways, merge away :)

PR #271, opened 2020-06-16
change for writing memory dataset from QGIS

vcloarec:
when writing data on faces with ASCII_DAT, if the uri doesn't contain "_els", MDAL change the uri. Then in QGIS, the dataset group is not retrieved because the uri has change.
With this PR, if the dataset group uri is changed if needed and exposed to the API to allows QGIS to register the good uri in `mExtraDatasetUris`.

For QGIS 3.14, not a big issue. The dataset is not loaded when first repoening, but the user can reload it, and everything will be ok after, even after close/repoen.

**EDIT:**
Add also the possibility to return file suffix for writing dataset

### DIFF
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -530,6 +530,12 @@ MDAL_EXPORT const char *MDAL_G_referenceTime( MDAL_DatasetGroupH group );
  */
 MDAL_EXPORT bool MDAL_G_isTemporal( MDAL_DatasetGroupH group );
 
+/**
+ * Returns dataset metadata key
+ * not thread-safe and valid only till next call
+ */
+MDAL_EXPORT const char *MDAL_G_uri( MDAL_DatasetGroupH group);
+
 ///////////////////////////////////////////////////////////////////////////////////////
 /// DATASETS
 ///////////////////////////////////////////////////////////////////////////////////////
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -164,6 +164,14 @@ MDAL_EXPORT bool MDAL_DR_meshLoadCapability( MDAL_DriverH driver );
 MDAL_EXPORT bool MDAL_DR_writeDatasetsCapability( MDAL_DriverH driver, MDAL_DataLocation location );
 
 /**
+ * Returns the file suffix used to write datasets on file
+ * not thread-safe and valid only till next call
+ *
+ * \since MDAL 0.7.0
+ */
+MDAL_EXPORT const char *MDAL_DR_writeDatasetsSuffix( MDAL_DriverH driver );
+
+/**
  * Returns whether driver has capability to save mesh
  */
 MDAL_EXPORT bool MDAL_DR_saveMeshCapability( MDAL_DriverH driver );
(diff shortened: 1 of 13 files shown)


mdal/api/mdal.h
```diff
@@ -530,6 +530,12 @@ MDAL_EXPORT const char *MDAL_G_referenceTime( MDAL_DatasetGroupH group );
  */
 MDAL_EXPORT bool MDAL_G_isTemporal( MDAL_DatasetGroupH group );
 
+/**
+ * Returns dataset metadata key
```
PeterPetrik, line 534: wrong doc

mdal/api/mdal.h
```diff
@@ -163,6 +163,14 @@ MDAL_EXPORT bool MDAL_DR_meshLoadCapability( MDAL_DriverH driver );
  */
 MDAL_EXPORT bool MDAL_DR_writeDatasetsCapability( MDAL_DriverH driver, MDAL_DataLocation location );
 
+/**
+ * Returns the file suffix used to write datasets on file
+ * not thread-safe and valid only till next call
+ *
+ * \since MDAL 0.7.0
+ */
+MDAL_EXPORT const char *MDAL_DR_writeDatasetsSuffix( MDAL_DriverH driver );
```
PeterPetrik, line 172: can you please point to the line where it is used in the QGIS PR? Not sure if I understand why it is needed.
- also it is not implemented for all drivers. (?)

vcloarec: here https://github.com/vcloarec/QGIS/pull/22/files#r443328274
and here : https://github.com/vcloarec/QGIS/pull/22/files#r443328741
Only implemented for driver that have write dataset capabilities, other : default value = "";

PeterPetrik: if we want to add it as a separate information, it should be good to follow how "filters" is implemented https://github.com/lutraconsulting/MDAL/blob/3f09e9f98e6533de24347213289e17daba13e919/mdal/frmts/mdal_driver.hpp#L81

so member variable that is initialized in the driver contructor.
-----
for the implementation, also we can just take mFilters variable, split by ;; and take the extension of the first enty? that case you do not need to add additional parameter to contructor etc.

PeterPetrik: maybe we can use filter alltogether? since when you save the file, you populate the filters in the save file dialog and user that way need to create file that fits the selected filters. so I am not sure we are not duplicating the infromataion?

PeterPetrik:
can you please fix the coverage, we want to be above 90%

mdal/api/mdal.h
vcloarec: I also thought that we could use the filter from the MDAL driver since I tried with Flo2d that have *.nc as a filter and save dataset with hdf5 format. So, as we can't purpose or force to use *.nc for hdf5, for this format, we may have a free suffix, not purpose *.nc. For ASCII_DAT and Binary DAT, I think it is better to force the file to have the suffix *.dat.
So after thought about, I think it is better, for each driver, to have filters that can select files that CAN be read, and, for drivers with write capabilities, we will have a suffix to force (or not if void) the written file. And also, IMHO, this gives more flexibility and clarity: filter to select existing files to read, suffix to force (or not) written files. As the base driver class have a default ("") for this suffix when writing, it is not a big deal to handle with final drivers. In the future, we will only have to implement the method if new drivers have the write capabilities.

Issue #287, opened 2020-08-03
Support for PLY


runette:
I know this has been addressed before - notably in #209.

However:

- in general, I am not sure I agree that PLY is not useful. It is a useful general format for meshes that hold scalar data since it makes absolutely no assumptions about the structure of the mesh or the data, however

- In the specific, I have a need to load PLY created by [PDAL](https://pdal.io). PDAL has a useful function for turning Pointcloud DEM to a mesh but the only usable format it can outpuit the mesh ini is PLY.

So I created a driver for PLY - ASCII only and Read-Only. Largely based on the XMS_TIN driver.

I will create a PR if you want to bring it into the project

PR #288, opened 2020-08-03
Add Driver for PLY format

runette:
see #287


mdal/frmts/mdal_ply.cpp
```diff
@@ -0,0 +1,445 @@
...
+  while ( true );
+
+  Vertices vertices( 0 );
+  Faces faces( 0 );
+  Edges edges( 0 );
+  size_t maxSizeFace = 0;
+  size_t faceSize = 0;
+
+  //datastructures that will contain all of the datasets, categorised by vertex, face and edge datasets
+  std::vector<std::vector<double>*> vertexDatasets; // conatins the data
```
vcloarec, line 189: why do you use pointer of vector<double>? Why not simply used `std::vector<std::vector<double>>?

mdal/frmts/mdal_ply.cpp
```diff
@@ -0,0 +1,445 @@
...
+        {
+          std::vector<double> &dataset = *vertexDatasets[j];
+          double value = MDAL::toDouble( chunks[MDAL::DriverPly::getIndex( element.properties, vProp2Ds[j] )] );
+          dataset.push_back( value );
+          vertexDatasets[j] = &dataset;
+        }
+      }
+
+      //
+      //load the trinagles
```
vcloarec, line 295: typo

mdal/frmts/mdal_ply.cpp
```diff
@@ -0,0 +1,445 @@
...
+      vertexCount = element.size;
+      vertices.resize( vertexCount );
+      for ( size_t i = 0; i < element.properties.size(); ++i )
+      {
+        if ( element.properties[i] != "x" &&
+             element.properties[i] != "y" &&
+             element.properties[i] != "z" &&
+             !element.list[i] )
+        {
+          vertexDatasets.push_back( new std::vector<double> );
```
vcloarec, line 225: where this std::vector<double> is deleted?

mdal/frmts/mdal_ply.cpp
```diff
@@ -0,0 +1,445 @@
...
+        Vertex &vertex = vertices[i];
+        vertex.x = MDAL::toDouble( chunks[MDAL::DriverPly::getIndex( element.properties, "x" )] );
+        vertex.y = MDAL::toDouble( chunks[MDAL::DriverPly::getIndex( element.properties, "y" )] );
+        vertex.z = MDAL::toDouble( chunks[MDAL::DriverPly::getIndex( element.properties, "z" )] );
+        for ( size_t j = 0; j < vProp2Ds.size(); ++j )
+        {
+          std::vector<double> &dataset = *vertexDatasets[j];
+          double value = MDAL::toDouble( chunks[MDAL::DriverPly::getIndex( element.properties, vProp2Ds[j] )] );
+          dataset.push_back( value );
+          vertexDatasets[j] = &dataset;
```
vcloarec, line 290: Not sure it is the good way to do. You replace a pointer to std::vector with another pointer without deleting the first one.
if you have `std::vector<std::vector<double>> vertexDatasets`
you can simply :
```
std::vector<double> &dataset = vertexDatasets[j];
double value = MDAL::toDouble( chunks[MDAL::DriverPly::getIndex( element.properties, vProp2Ds[j] )] );
dataset.push_back( value );
```

or
```
double value = MDAL::toDouble( chunks[MDAL::DriverPly::getIndex( element.properties, vProp2Ds[j] )] );
vertexDatasets[j].push_back( value );
```

mdal/frmts/mdal_ply.cpp
runette: Mostly because with all of the later updates coming to both dimensions of the array it felt more efficient to just hold the pointer.

But to be honest, I don't do a lot of cpp, So if you felt it to be better I could change it - it could make things a little simpler later on.

mdal/frmts/mdal_ply.cpp
runette: Oops. Fixed in my source. It will get upload in the next commit.

mdal/frmts/mdal_ply.cpp
runette: The first was the original intent but it does not work.

The second might work and be clearer. I will try it.

vcloarec: > The first was the original intent but it does not work.

If you use `std::vector<std::vector<double>> vertexDatasets` without pointer, I don't see any reason why it doesn't work

mdal/frmts/mdal_ply.hpp
```diff
@@ -0,0 +1,59 @@
...
+      struct element
+      {
+        public:
+          std::string name; // element name
+          std::vector<std::string> properties; // the name of each property
+          std::vector<std::string> types; // the type of each property
+          std::vector<bool> list; // is the property a list
+          size_t size; // element size
+
+          bool operator==( const std::string rhs )
```
vcloarec, line 49: `const std::string &rhs`

mdal/frmts/mdal_ply.cpp
```diff
@@ -0,0 +1,445 @@
...
+      {
+        if ( element.properties[i] != "vertex_indices" &&
+             !element.list[i] )
+        {
+          faceDatasets.push_back( new std::vector<double> );
+          fProp2Ds.push_back( element.properties[i] );
+        }
+      }
+    }
+    else if ( element.name == "edge" )
```
vcloarec, line 244: I am not sure MDAL can support, for now, having faces and edges together. Edge is, for now, used for 1D Mesh, that is network of node linked only by edges.

mdal/frmts/mdal_ply.hpp
```diff
@@ -0,0 +1,59 @@
+/*
+ MDAL - Mesh Data Abstraction Library (MIT License)
+ Copyright (C) 2020 Peter Petrik (zilolv at gmail dot com)
```
PeterPetrik, line 3: Probably you want to own the copyright yourself :)

mdal/frmts/mdal_ply.hpp
```diff
@@ -0,0 +1,59 @@
...
+          std::vector<bool> list; // is the property a list
+          size_t size; // element size
+
+          bool operator==( const std::string rhs )
+          {
+            return name == rhs;
+          }
+      };
+
+      size_t getIndex( std::vector<std::string> v, std::string in );
```
PeterPetrik, line 55: you probably want cont & for both args here?

tests/test_ply.cpp
```diff
@@ -0,0 +1,365 @@
+/*
+ MDAL - Mesh Data Abstraction Library (MIT License)
+ Copyright (C) 2018 Peter Petrik (zilolv at gmail dot com)
```
PeterPetrik, line 3: Probably you want to own the copyright yourself :)

mdal/frmts/mdal_ply.cpp
```diff
@@ -0,0 +1,445 @@
+/*
+ MDAL - Mesh Data Abstraction Library (MIT License)
+ Copyright (C) 2020 Lutra Consulting Ltd.
```
PeterPetrik, line 3: Probably you want to own the copyright yourself :)

mdal/frmts/mdal_ply.cpp
```diff
@@ -0,0 +1,445 @@
...
+MDAL::DriverPly *MDAL::DriverPly::create()
+{
+  return new DriverPly();
+}
+
+MDAL::DriverPly::~DriverPly() = default;
+
+size_t MDAL::DriverPly::getIndex( std::vector<std::string> v, std::string in )
+{
+  auto it = find( v.begin(), v.end(), in );
```
PeterPetrik, line 47: please do not use auto

mdal/frmts/mdal_ply.cpp
```diff
@@ -0,0 +1,445 @@
...
+  std::vector<MDAL::DriverPly::element> elements; // we will decant the element specifications into here
+  std::string proj = "";
+
+  if ( !std::getline( in, line ) )
+  {
+    MDAL::Log::error( MDAL_Status::Err_IncompatibleMesh, name(), meshFile + " Header is corrupt" );
+    return nullptr;
+  }
+
+  //
```
PeterPetrik, line 84: please use /* */ style for multi-level comments

mdal/frmts/mdal_ply.cpp
runette: I have three comments to make :

1) The PLY format and this driver are very flexible about the data contents. This is not to say that there MUST be both edges and faces. Just that neither the format nor the driver enforce that there is NOT both edges and faces, 

2) As the tests show, MDAL can support data that has both faces and edges. In the test data I am loading and then successfully reading datasets that contain data on both the faces and edges. That data has no particular GIS meaning , but MDAL processes it correct;y as I understand that it should. It is up to the host program if that data makes sense, and

3) I think that you are saying that it is possible to load data from a PLY file that does not make sense to other formats and cannot be converted to other formats successfully. Isn't that in the nature of a DAL? You either end up enforcing a lowest common denominator type of entity model and have no effective value or you accept that the host application and the user needs to apply to intelligence about what data can convert to what.

PeterPetrik: You are right @runette . What Vincent tried to express that the "host" program QGIS now, and other formats have either 1D or 2D data (so for example we read it as 2 separate meshes usually, one for 1D data). Such combined mesh is fine, but there could be some bugs in QGIS displaying that since noone tried it yet.

Best to check in QGIS to see if it is supported and possibly report bugs

mdal/frmts/mdal_ply.hpp
runette: Done in the next commit

tests/test_ply.cpp
runette: Done in the next commit

mdal/frmts/mdal_ply.cpp
runette: Done in the next commit

mdal/frmts/mdal_ply.cpp
runette: Probably easier and more effective to just put a note in the driver description. I cannot think of a real use-case that uses both - certainly for QGIS data although there may be for what I am using MDAL for. If a use-case for data both on faces and edges in QGIS comes up then the use-case owner can raise that with QGIS.

runette: Done in the next commit

runette:
I made some changes to address the issues raised so far

PeterPetrik:
@runette for style you can just run https://github.com/lutraconsulting/MDAL/blob/master/scripts/check_all.bash locally

PeterPetrik (approved):

PeterPetrik:
@runette thank you for the contribution, well done!

runette:
I have never got it to work ... :( Probably because my dev env is Windows

PR #292, opened 2020-08-16
Mesh creation

vcloarec:
This PR allows the empty mesh creation and the possibility to add vertices and faces to existing mesh.
Linked with https://github.com/qgis/QGIS/pull/38309

### DIFF
--- a/tests/test_selafin.cpp
+++ b/tests/test_selafin.cpp
@@ -115,6 +115,10 @@ TEST( MeshSLFTest, MalpassetGeometry )
   value = getValue( ds, 9571 );
   EXPECT_DOUBLE_EQ( 7.5623664855957, value );
 
+  std::vector<double> newVertex{10, 10, 10};
+  MDAL_M_addVertices( m, 1, newVertex.data() );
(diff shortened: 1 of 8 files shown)


tests/test_selafin.cpp
```diff
@@ -115,6 +115,10 @@ TEST( MeshSLFTest, MalpassetGeometry )
   value = getValue( ds, 9571 );
   EXPECT_DOUBLE_EQ( 7.5623664855957, value );
 
+  std::vector<double> newVertex{10, 10, 10};
+  MDAL_M_addVertices( m, 1, newVertex.data() );
+  EXPECT_EQ( MDAL_LastStatus(), Err_IncompatibleMesh );
```
PeterPetrik, line 120: add test that selafin mesh is not editable too,.

vcloarec: hmm, this line tests that selafin mesh is not editable, I am not sure to understand. Do you want I add a `MDAL_Status` for non editable mesh?

PeterPetrik: no problem

PeterPetrik (approved):

PR #305, opened 2020-10-20
Fix for #304

### LINKED ISSUE #304: mdal_api_test fails if MDAL is built without HDF5
ERROR

You can build MDAL without HDF5 using the -DWITH_HDF5=OFF switch and the build works, but the tests fail because mdal_api_test expects the build to have used HDF5

RESULT

ApiTest.MeshNamesApi fails

EXPECTED RESULT

All tests should be passed 

PROPOSED ACTION

I think that test_api.cpp probably needs to be split into two - one for tests that do not need HDF5 and one for tests that need it.

runette:
Proposal to add a compiler directive to tell if HDF5 is present and use this to determine what tests to run.

closes #304

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,6 +1,12 @@
 # MDAL - Mesh Data Abstraction Library (MIT License)
 # Copyright (C) 2018 Peter Petrik (zilolv at gmail dot com)
 
+IF (HDF5_FOUND)
+  ADD_COMPILE_DEFINITIONS(HDF5=1)
+ELSE (HDF5_FOUND)
+
+ENDIF (HDF5_FOUND)
+
 FIND_PACKAGE(Threads REQUIRED)
 FIND_PACKAGE(GTest)
 
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,12 +1,6 @@
 # MDAL - Mesh Data Abstraction Library (MIT License)
 # Copyright (C) 2018 Peter Petrik (zilolv at gmail dot com)
 
-IF (HDF5_FOUND)
-  ADD_COMPILE_DEFINITIONS(HDF5=1)
-ELSE (HDF5_FOUND)
-
-ENDIF (HDF5_FOUND)
-
 FIND_PACKAGE(Threads REQUIRED)
 FIND_PACKAGE(GTest)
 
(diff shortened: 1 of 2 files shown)


tests/test_api.cpp
```diff
@@ -333,16 +325,33 @@ TEST( ApiTest, MeshNamesApi )
...
-      "\"\"" + test_file( "/flo2d/BarnHDF5/TIMDEP.HDF5" ) + "\"\"",
-      "FLO2D:\"" + test_file( "/flo2d/BarnHDF5/TIMDEP.HDF5" ) + "\":" + "mesh2d"
-    },
-    {
-      "2DM:\"" + test_file( "/flo2d/BarnHDF5/TIMDEP.HDF5" ) + "\"",
-      "2DM:\"" + test_file( "/flo2d/BarnHDF5/TIMDEP.HDF5" ) + "\""
-    },
   };
-
+#if HDF5 == 1
```
PeterPetrik, line 329: see https://github.com/lutraconsulting/MDAL/blob/8ea01b603b21d929c61c4b5d0f85830099011b0f/mdal/mdal_driver_manager.cpp#L210

tests/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 # MDAL - Mesh Data Abstraction Library (MIT License)
 # Copyright (C) 2018 Peter Petrik (zilolv at gmail dot com)
 
+IF (HDF5_FOUND)
```
PeterPetrik, line 4: this is not required, the compiler definitions are already generated by https://github.com/lutraconsulting/MDAL/blob/8ea01b603b21d929c61c4b5d0f85830099011b0f/CMakeLists.txt#L103 in file mdal_config.hpp 

It is enough in the test to include this file and use macro HAVE_HDF5

runette: Using HDF5_FOUND  was my first idea but it did not work - it alwas thought the HDF5 was included. Thus the slight cludge.

But - I had not spotted the use of ifdef HAVE_HDF5. I will try that and see if I can get it to work.

PeterPetrik: one is in the cmake , other in c++ files

runette: Ok - that is done and looks like it is working

PeterPetrik (approved):

PR #319, opened 2020-11-20
dynamic driver

vcloarec:
Allow the possibility to add external lib file to be loaded by MDAL.


tools/externaldriver/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+SET(MDAL_DUMMY_EXTERNAL_DRIVER
+    mdalDummyExternalDriver.cpp)
+	
+INCLUDE_DIRECTORIES(
+    ${CMAKE_CURRENT_SOURCE_DIR}
+    ${CMAKE_CURRENT_BINARY_DIR}
+    ${CMAKE_BINARY_DIR}
+    ${CMAKE_SOURCE_DIR}
+    ${CMAKE_SOURCE_DIR}
+)
```
PeterPetrik, line 10: not sure if you need this at all. you need binary dirs only if you generate some headers, maybe at maximum ${CMAKE_SOURCE_DIR}

tools/externaldriver/mdalDummyExternalDriver.cpp
```diff
@@ -0,0 +1,598 @@
...
+    const Datasetgroup &datasetGroup = mesh.datasetGroups.at( groupIndex );
+    if ( datasetGroup.dataType == "onFace" )
+      return true;
+  }
+
+  return false;
+}
+
+//! Returns the dataset active flags
+int CALL_CONV  MDAL_DRIVER_D_activeFlags( int meshId, int groupIndex, int datasetIndex, int indexStart, int count, int *buffer )
```
PeterPetrik, line 570: i do not think you need to repeat CALL_CONV in the implementation? also i have never seen speciying calling conventions... is it reequired at all? (looks like the one specified is default)

we do not specify it for example in the main MDAL header too

tools/externaldriver/mdal_external_driver.h
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Returns edges from iterator for the mesh
+ * \param meshId id of the requested mesh
+ * \param edgeCount maximum number or edges to be written to buffer
+ * \param startVertexIndices must be allocated to edgesCount items to store start vertex indices for edges
+ * \param endVertexIndices must be allocated to edgesCount items to store end vertex indices for edges
+ * \returns number of vertices effectivly written in the buffer
+ *
+ * \since MDAL 0.6.0
```
PeterPetrik, line 105: remove since

tools/externaldriver/mdal_external_driver.h
```diff
@@ -0,0 +1,143 @@
...
+#      define MDAL_LIB_EXPORT
+#    endif
+#  endif //_WIN32 || defined __CYGWIN__
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+//! Returns the driver name
```
PeterPetrik, line 34: can you double check the comments here?

PeterPetrik (approved):
wonderful

mdal/frmts/mdal_dynamic_driver.cpp
```diff
@@ -0,0 +1,437 @@
...
+    return nullptr;
+  }
+
+  return driver.release();
+}
+
+bool MDAL::DriverDynamic::loadSymbols()
+{
+  mCanReadMeshFunction = mLibrary.getSymbol<bool, const char *>( "MDAL_DRIVER_canReadMesh" );
+  mOpenMeshFunction = mLibrary.getSymbol<int, const char *, const char *>( "MDAL_DRIVER_openMesh" );
```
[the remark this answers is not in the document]
vcloarec: yes, it is a bit tricky...
the driver created with `MDAL::DriverDynamic::create( const std::string &libFile )` (line 66/98) is the driver stored by the driver manager. This driver will have to create the drivers which will load the mesh. The first one need only the functions loaded on line 70/74).
The driver which will load the mesh, need only functions loaded on line 105/106.

But agree, it is quite confusing... I will put all the loading on the same place.

tools/externaldriver/mdalDummyExternalDriver.cpp
vcloarec: I put this when trying to make working on windows... I tried without, it seems ok. So I will remove it.

mdal/frmts/mdal_dynamic_driver.cpp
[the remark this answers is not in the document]
vcloarec: Finally, the first series of functions (in M`DAL::DriverDynamic::create( const std::string &libFile )`) are only needed when loading the driver library to create the driver in the driver manager. This function are not needed internally in the driver after loading it. So these first function are only used when the driver does not exist yet.
Once library loaded and driver created, the driver needs only `mCanReadMeshFunction` and `mOpenMeshFunction` that are member of the driver class.

PR #320, opened 2020-11-23
fix hecras velocity/shear stress

vcloarec:
Affter playing a bit with hecras model and loading in QGIS, I've seen that velocity was not the same in hecras and QGIS.
Indeed, in hdf file, the velocity (and sheer stress) is quite tricky : the values are the values of vector (velocity or shear stress) projected on the normal of faces (that are  edges of the mdal faces). Confusing and not the same as the real value...
So, the real values need to reconstructed to obtain a vector value.

This PR allows to reconstruct the real values as vector, so hecras format support now velocity (and shear stress) as vector value.

I think it could be good to backport this to 0.7.1 to have this fix with the future LTR of QGIS.

### DIFF
--- a/mdal/frmts/mdal_hec2d.cpp
+++ b/mdal/frmts/mdal_hec2d.cpp
@@ -202,21 +202,6 @@ static std::vector<MDAL::RelativeTimestamp> readTimes( const HdfFile &hdfFile )
   return convertTimeData( times, dataTimeUnits );
 }
 
-static std::vector<int> readFace2Cells( const HdfFile &hdfFile, const std::string &flowAreaName, size_t *nFaces )
-{
-  // First read face to node mapping
-  HdfGroup gGeom = openHdfGroup( hdfFile, "Geometry" );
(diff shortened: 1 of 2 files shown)


PeterPetrik:
Thanks for fixing this. Would it be possible to update documentation in the header of the driver?

PeterPetrik (approved):

vcloarec:
what about backport?

PeterPetrik:
I do not know, the bug is not crashing anything and no client asked for it and in my opinion it is more of enhancement than bugfix. I would backport it when someone ask for it.

vcloarec:
Hec-Ras is a very used software... actually, the velocity values are wrong and that could make go away many users.

PeterPetrik:
@saberraz, do you want to backport this to qgis 3.16?

saberraz:
+1 to backport!

PeterPetrik:
@vcloarec can you create backport PR to branch release_0_7 ? thanks

Issue #218, opened 2020-02-20
Support 2DM's E6T element (triangle with 6 nodes)


Brent-Edwards:
I tried to create the simplest possible example for test purposes - see [http://forum.aquaveo.com/topic/3426-working-example-of-a-2dm-mesh-file-with-e6t-elements-six-node-quadratic-triangles/](http://forum.aquaveo.com/topic/3426-working-example-of-a-2dm-mesh-file-with-e6t-elements-six-node-quadratic-triangles/)

Unfortunately, I haven't received any responses from the SMS forum.

In QGIS, the file seems to read fine, but doesn't render:
![image](https://user-images.githubusercontent.com/13538726/74962909-cdded780-53de-11ea-850c-917c4e60ad74.png)

Crayfish exports the nodes, but produces an empty (nodata) raster and empty polygon for the face (expression used to create the "wkt" virtual field -  geom_to_wkt( $geometry) ) : ![image](https://user-images.githubusercontent.com/13538726/74963081-1eeecb80-53df-11ea-8795-30d1b6342d85.png)
```
QGIS version info:


QGIS version
3.10.2-A Coruña
QGIS code revision
d4cd3cfe5a
Compiled against Qt
5.11.2
Running against Qt
5.11.2
Compiled against GDAL/OGR
3.0.3
Running against GDAL/OGR
3.0.3
Compiled against GEOS
3.8.0-CAPI-1.13.1
Running against GEOS
3.8.0-CAPI-1.13.1 
Compiled against SQLite
3.29.0
Running against SQLite
3.29.0
PostgreSQL Client Version
11.5
SpatiaLite Version
4.3.0
QWT Version
6.1.3
QScintilla2 Version
2.10.8
Compiled against PROJ
6.3.0
Running against PROJ
Rel. 6.3.0, January 1st, 2020
OS Version
Windows 10 (10.0)
Active python plugins
changeDataSource; 
crayfish; 
gmsh; 
latlontools; 
Mergin; 
mmqgis; 
profiletool; 
QMetaTiles; 
quick_map_services; 
shapetools; 
splitmultipart; 
SVG2ColoR; 
timemanager; 
valuetool; 
VectorMCDA; 
VoGisProfilTool; 
db_manager; 
MetaSearch; 
processing

```

wonder-sk:
MDAL currently does not support E6T elements. Only E3T and E4Q elements are supported.

[...]

PR #324, opened 2020-12-01
2dm enhancements

PeterPetrik:


PeterPetrik:
fix #325 

fix #218

mdal/frmts/mdal_2dm.cpp
```diff
@@ -50,6 +50,21 @@ bool _parse_vertex_id_gaps( std::map<size_t, size_t> &vertexIDtoIndex, size_t ve
   return false;
 }
 
+void _persist_native_index( std::vector<double> &arr, size_t nativeID, size_t ourId, size_t maxOurId )
```
vcloarec, line 53: static ?

vcloarec (commented):
looks good.
If there is no gap between native and mdal indexes, the indexes are not exposed as dataset group? In QGIS, how the user can see the indexes?

PeterPetrik:
> If there is no gap between native and mdal indexes, the indexes are not exposed as dataset group? In QGIS, how the user can see the indexes?

i think if there are no gaps, it is simple (feature ID - 1)

PR #334, opened 2021-01-10
hecras: fix version 6.0 and wet/dry faces

### LINKED ISSUE #333: Mesh Hec Ras 6.0
New version of Hec Ras (6.0),  new hdf file doesn't work in Qgis

[Prueba.p01.zip](https://github.com/lutraconsulting/MDAL/files/5736316/Prueba.p01.zip)

vcloarec:
fix #333 
New format  is very similar but the dataset group are not exactly the same (Depth and shear stress are now optional)

There are also optional dataset groups that are not yet supported by MDAL.

This PR fix also the comparison for wet/dry faces.

### DIFF
--- a/mdal/frmts/mdal_hec2d.cpp
+++ b/mdal/frmts/mdal_hec2d.cpp
@@ -237,7 +237,16 @@ void MDAL::DriverHec2D::readFaceOutput( const HdfFile &hdfFile,
     std::string flowAreaName = flowAreaNames[nArea];
 
     HdfGroup gFlowAreaRes = openHdfGroup( rootGroup, flowAreaName );
-    HdfDataset dsVals = openHdfDataset( gFlowAreaRes, rawDatasetName );
+    HdfDataset dsVals;
+    try
+    {
+      dsVals = openHdfDataset( gFlowAreaRes, rawDatasetName );
+    }
+    catch ( MDAL::Error )
+    {
+      return;
+    }
+
     std::vector<float> vals = dsVals.readArray();
 
     HdfGroup gGeom = openHdfGroup( hdfFile, "Geometry" );
@@ -334,7 +343,7 @@ void MDAL::DriverHec2D::readFaceOutput( const HdfFile &hdfFile,
     }
   }
 
-  for ( auto dataset : datasets )
+  for ( auto &dataset : datasets )
   {
     dataset->setStatistics( MDAL::calculateStatistics( dataset ) );
     group->datasets.push_back( dataset );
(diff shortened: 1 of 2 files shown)


mdal/frmts/mdal_hec2d.cpp
```diff
@@ -398,7 +407,16 @@ std::shared_ptr<MDAL::MemoryDataset2D> MDAL::DriverHec2D::readElemOutput( const
...
 
-    HdfDataset dsVals = openHdfDataset( gFlowAreaRes, rawDatasetName );
+    HdfDataset dsVals;
+    try
+    {
+      dsVals = openHdfDataset( gFlowAreaRes, rawDatasetName );
+    }
+    catch ( MDAL::Error )
+    {
+      return nullptr;
```
PeterPetrik, line 417: function returns void

mdal/frmts/mdal_hec2d.cpp
```diff
@@ -431,7 +449,7 @@ std::shared_ptr<MDAL::MemoryDataset2D> MDAL::DriverHec2D::readElemOutput( const
             {
               assert( bed_elevation );
               double bed_elev = bed_elevation->scalarValue( eInx );
-              if ( std::isnan( bed_elev ) || fabs( bed_elev - val ) > eps ) // change from bed elevation
+              if ( std::isnan( bed_elev ) || fabs( val - bed_elev ) > ( val + bed_elev )*eps )  // change from bed elevation
```
PeterPetrik, line 452: not sure if I understand this comparison

tests/test_hec2d.cpp
```diff
@@ -392,8 +392,8 @@ TEST( MeshHec2dTest, model_505 )
   count = MDAL_D_valueCount( ds );
   ASSERT_EQ( 73, count );
 
-  value = getValue( ds, 10 );
-  EXPECT_DOUBLE_EQ( 33.000003814697266, value );
+  value = getValue( ds, 50 );
+  EXPECT_TRUE( MDAL::equals( 34.785, value, 0.001 ) );
```
PeterPetrik, line 396: why the value changed?

mdal/frmts/mdal_hec2d.cpp
vcloarec: epsilon is the difference between 1.0 and the next value representable by the floating-point

For a float epsilon # 0.000000X 
To compare float (7 signifiant digits) that are not next to 1.0, need to transform epsilon to give it the order of magnitde of the compared float. 
Several way can be used depending on context, here the sum is enought.

mdal/frmts/mdal_hec2d.cpp
vcloarec: std::shared_ptr<MDAL::MemoryDataset2D> is not void.
but make think it will be better to throw a exception outside this method if this is bed elevation

tests/test_hec2d.cpp
vcloarec: with dry/wet fixed, getValue( ds, 10 ) = NaN, 
but better to keep the test on NaN and add a test to not NaN

PeterPetrik (approved):

vcloarec:
backport to 0.7? Same reasons than #320

PeterPetrik:
@saberraz ?

saberraz:
Why not?! :)

PR #353, opened 2021-05-12
fix #347

### LINKED ISSUE #347: Import of NetCDF fails upon order of variables
see https://github.com/qgis/QGIS/issues/42552

@raphaeljolivet

vcloarec:
fix the QGIS issue https://github.com/qgis/QGIS/issues/42552


mdal/frmts/mdal_gdal.cpp
```diff
@@ -564,36 +564,76 @@ std::unique_ptr<MDAL::Mesh> MDAL::DriverGdal::load( const std::string &fileName,
...
-        // If it is first dataset, create mesh from it
+        firstProjFound = true;
         gdal_datasets.push_back( cfGDALDataset );
+      }
+      else
+      {
+        datasets.push_back( cfGDALDataset );
+      }
+    }
+
```
PeterPetrik, line 589: here I would just add 
```
if !firstProjFound
    take first one
```

and then the rest of the function can be the same for both cases?

vcloarec: or last commit?

PeterPetrik (approved):

PR #362, opened 2021-06-23
allow void mesh for mesh writing

vcloarec:
In the scope of mesh editing in QGIS, when creating a new mesh, QGIS need to write a empty mesh. With the actual implementation of MDAL for  UGRID and SELAFIN format, there some issues.
This PR allows the possibility to have create void mesh.
For UGRID, the problem is for dimension the value of 0 is considered as NC_UNLIMITED, that is no limit for the dimension. And for netCDF, it is allow for only one dimension... as we need for vertices, faces, edges, time step and eventually max vertices per face, we can't use this NC_UNLIMITED. The workaround is to use the value 1 for dimension and fake vertices/faces. When reading those fake elements are detected and removed.


tests/test_ugrid.cpp
```diff
@@ -704,6 +704,29 @@ TEST( MeshUgridTest, magnitude_direction )
...
+
+  EXPECT_EQ( MDAL_MeshNames( uriToMeshNames.c_str() ), "Ugrid:\"" + path + "\":mesh2d" );
+  MDAL_MeshH m = MDAL_LoadMesh( uri.c_str() );
+
+  ASSERT_NE( m, nullptr );
+  MDAL_Status s = MDAL_LastStatus();
+  EXPECT_EQ( MDAL_Status::None, s );
+
+  int v_count = MDAL_M_vertexCount( m );
+  EXPECT_EQ( v_count, 0 );
```
PeterPetrik, line 722: I have a question, if I have void mesh and want to add a single face (I would like to see in the auto-tests to see that we have f_count=1 afterwards), I need to RECREATE the whole file from scratch or I can just "update" it??

I think in general we want to have a new capability on driver if they are able to update the mesh too? I imagin for netcdf-based formats it should be possible.?. and it would be kind of nice to not need to recreate it from scratch?

vcloarec: For any update, you need to rewrite the whole file. It seems difficult for now to allow updating only parts of the file for several reasons.
But indeed, it will be a good new capability and it will be possible with the netcdf based format. But, I think it will need quite big changes on the way the mesh is handle for this format For, now it is a memory mesh, so for updating only some parts we need to sync the memory mesh with the mesh in the file.

PR #364, opened 2021-06-25
Add Status calls to C API

### LINKED ISSUE #359: Add the ability to  change the MDAL status in the C API
As part of creating the Python wrapper for the C API, I have noticed some things missing from the C API.

One of them is the ability to raise error, warning or debug status messages into MDAL using the API.

Basically, I have a C++ wrapper that marshalls the basic mesh object into numpy arrays for python. That wrapper needs to be able to raise errors and the best way to do that - so that all non python status messages come into python the same way - would be to set the MDAL Last Status. However, the C++ calls to do that are not exposed in the C API.

Raising it here as an enhancement request and for comment. I will probably try to raise an accompanying PR in the next week if I can.

runette:
fixes #359 

Adds two calls to the C API 

- void MDAL_ResetStatus();
- void MDAL_SetStatus( MDAL_LogLevel Level, MDAL_Status status, const char *message );

### DIFF
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -112,6 +112,17 @@ MDAL_EXPORT const char *MDAL_Version();
 MDAL_EXPORT MDAL_Status MDAL_LastStatus();
 
 /**
+ * Resets the last status message
+ */
+MDAL_EXPORT void MDAL_ResetStatus();
+
+/*
+ * Sets the last status message
+ */
+
+MDAL_EXPORT void MDAL_SetStatus( MDAL_LogLevel Level, MDAL_Status status, const char* message);
+
+/**
  * Sets custom callback for logging output
  *
  * By default standard stdout is used as output.
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -120,7 +120,7 @@ MDAL_EXPORT void MDAL_ResetStatus();
  * Sets the last status message
  */
 
-MDAL_EXPORT void MDAL_SetStatus( MDAL_LogLevel Level, MDAL_Status status, const char* message);
+MDAL_EXPORT void MDAL_SetStatus( MDAL_LogLevel Level, MDAL_Status status, const char *message );
 
 /**
  * Sets custom callback for logging output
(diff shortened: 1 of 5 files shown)


PeterPetrik:
@runette I am a bit unsure in general it is relevant or necessary. My understanding of the MDAL_LastStatus is more about how to notify about error in the last call of any MDAL function. It doesn't seems right to push the error message to the MDAL library for error that happened outside the library? Can you point me the the python wrapper implementation where it is required? On the other hand, imagine you are user of the MDAL library and you find out that MDAL_lastStatus is not the last status of MDAL API call but some status someone pushed from outside... that could be confusing.

runette:
@PeterPetrik The basic example is the [python wrapper e.g. DatasetGroup](https://github.com/ViRGIS-Team/mdal-python/blob/e0d888b216771d2ee23904d5b7eae83d1089e9c8/mdal/DatasetGroup.cpp#L85). This is, of course, my dev version that I used for testing the PR and not the currently live version.

The wrapper is needed to marshall the data between numpy and MDAL and is the same basic idea as the PDAL wrapper (which I copied). There are a number of errors that can happen - both in the MDAL and in the wrapper - and doing it this way means that the calling function only needs to perform one check for status - [see the wrapper code](https://github.com/ViRGIS-Team/mdal-python/blob/383d3c73cacee37b1765063135570d393229f051/mdal/libmdalpython.pyx#L646).

I see the ResetStatus call as vital since MDAL itself is a bit untidy about when the status is reset - I have been getting problems checking the status since certain calls do not set it ( save was one ). Not being able to reset the status from the C API is a problem.

The SetStatus call is good to have - for the reason described above. Otherwise, the caller needs to implement two different status calls and check both.

I am also not sure about the objection. The call only updates the current instance and by definition, if there is a wrapper the user is accessing that MDAL instance through the wrapper. I am not sure why the user would be surprised to see errors raised by the wrapper and they are also by definition only seeing the MDAL errors if they are forwarded through the wrapper. To me this seems logical and consistent.

PeterPetrik:
ok, not a big fan, but also do not see a problem of having it. lets merge and see how it evolves. Thanks for the efforts with the python api!

PR #367, opened 2021-07-09
Add Metadata to Mesh

runette:
Working on making the Python interface write enabled and making the PLY driver write-enabled, I have realised that there is a need for metadata on the mesh. The primary use case for this is to retain format related information (in the PLY case e.g. whether the file is text or binary) to allow round trip consistency.

My preferred way to do this is to add the same metadata capabilities to the mesh object as is currently found on the DatasetGroup object (rather than try to proliferate driver specific functionality).

I am proposing that we basically copy the metadata related methods and functionality from the DatasetGroup object to the Mesh object to provide consistency.

This is a NON BREAKING change to the API - since it does not change any current functionality and does not require that the driver uses the metadata in any way. It will, however, allow more flexibility for future drivers etc.

### DIFF
--- a/mdal/mdal_data_model.hpp
+++ b/mdal/mdal_data_model.hpp
@@ -267,6 +267,12 @@ namespace MDAL
       std::string crs() const;
       size_t faceVerticesMaximumCount() const;
 
+      std::string getMetadata( const std::string &key );
+      void setMetadata( const std::string &key, const std::string &val );
+      void setMetadata( const Metadata &metadata );
+
+      Metadata metadata;
+
       virtual void closeSource() {};
 
       virtual bool isEditable() const {return false;}
--- a/mdal/mdal_data_model.hpp
+++ b/mdal/mdal_data_model.hpp
@@ -152,7 +152,7 @@ namespace MDAL
 
       std::string getMetadata( const std::string &key );
       void setMetadata( const std::string &key, const std::string &val );
-      void setMetadata( const Metadata &metadata );
+      void setMetadata( const Metadata &new_metadata );
 
       std::string name();
       void setName( const std::string &name );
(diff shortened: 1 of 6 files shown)


PeterPetrik (approved):

mdal/mdal_data_model.hpp
```diff
@@ -152,7 +152,7 @@ namespace MDAL
 
       std::string getMetadata( const std::string &key );
       void setMetadata( const std::string &key, const std::string &val );
-      void setMetadata( const Metadata &metadata );
+      void setMetadata( const Metadata &new_metadata );
```
PeterPetrik, line 155: why the change of the arg name?

mdal/mdal_data_model.hpp
```diff
@@ -267,6 +267,12 @@ namespace MDAL
       std::string crs() const;
       size_t faceVerticesMaximumCount() const;
 
+      std::string getMetadata( const std::string &key );
```
PeterPetrik, line 270: we do not use getXYZ() syntax for getters, can you rename to `std::string metadata(...)` ?

mdal/mdal_data_model.hpp
runette: because having the same name provokes a compiler warning in windows since the variable hides a class member.

Not a big problem - but it does make sense to avoid the compiler warnings.

mdal/mdal_data_model.hpp
runette: I actually copied the structure and names exactly from DatasetGroup - including the getter - for complete consistency.

I cannot rename to `metadata` - because there is already a class member Mesh::metadata (and DatasetGroup::metadata since the former is an exact copy of the latter in structure).

The `getMetadata()` method does add the ability to get metadata by key as opposed to by index. In the DatasetGroup case this is used in one place  DAT driver.

I can rename the method anything you want (except metadata) but I do strongly suggest that we keep the signatures for Mesh metadata and DatasetGroup metadata consistent, otherwise we are just creating work and potential errors.

PeterPetrik: ok

PR #368, opened 2021-07-11
Write enabled and Binary enabled PLY driver

runette:
This PR is the first part of an update to the PLY driver to make to READ/WRITE and to include binary as well as ASCII format files. Currently, although it does both formats it has only been tested on ASCII formats. There will be an update when the binary formats are tested and working (which will be dependent on the metadata PR submitted separately). Submitting now to start the review.

NOTES:

1 This driver has been built on top of an existing library for PLY. This code is fully MIT licenced (include explicit language allowing the code to be merged) and I have included the copyright notice both in the MDAL LICENCE.TXT and in its own licence file. I believe this meets all legal requirements. Note that the libplyxx code has been extensively customised to the needs of MDAL, has been changed to use MDAL error logging and has been updated to include all PLY data types and lists in all positions.

2 I have build the PLY library code into the MDAL shared object to avoid problems with creating new dependencies. This may be seen as not usual - but the amount of code involved is no more than would be required to write a PLY driver from scratch and it does not bring in any unusual or new dependencies itself (except as in the next point) - so this seems reasonable to me. 

3 This code DOES make extensive use of std::unordered_map and therefore has a dependency on C++ 14. This does not seem unreasonable in 2021.

4 The use of libplyxx approach means that the actual mdal_ply code is such simplified down to just wrangling the MDAL into and out of the libplyxx data formats.


mdal/frmts/mdal_ply.cpp
```diff
@@ -60,275 +65,190 @@ bool MDAL::DriverPly::canReadMesh( const std::string &uri )
...
+           property.name != "Y" &&
+           property.name != "y" &&
+           property.name != "Z" &&
+           property.name != "z"
+         )
       {
-        MDAL::Log::error( MDAL_Status::Err_IncompatibleMesh, name(), meshFile + " contains invalid line : " + line );
-        return nullptr;
+        vProp2Ds.push_back( property.name );
+        vertexDatasets.push_back( * new std::vector<double> );
```
vcloarec, line 123: I don't look at all the code, I fine this line odd.
Can you explain the logic of this? When this `std::vector<double>` will be deleted?
Maybe I am wrong, but for me here, you create a `std::vector` with a pointer to it, you de-reference the pointer with `*`, to copy the pointer object in the `vertexDatasets` container. The first created std::vector is never deleted.
Why do you not use `vertexDatasets.push_back(std::vector<double>() );`

mdal/frmts/mdal_ply.cpp
```diff
@@ -60,275 +65,190 @@ bool MDAL::DriverPly::canReadMesh( const std::string &uri )
...
+          else if ( p.name == "Z" || p.name == "z" )
+          {
+            vertex.z = e[i];
+          }
+          else
+          {
+            int dsIdx = getIndex( vProp2Ds, p.name );
+            std::vector<double> *ds = & vertexDatasets[dsIdx];
+            ds->push_back( e[i] );
+          }
```
vcloarec, line 170:
```suggestion
            std::vector<double> &ds = vertexDatasets[dsIdx];
            ds.push_back( e[i] );
          }
```

mdal/frmts/mdal_ply.cpp
runette: Despite the impression given by the diff - those lines are not changed from the previous version! 

However, since I wrote the previous version as well ...

I think you are right - I am not sure that if I were writing it now I would like that. I will have a look.

mdal/frmts/mdal_ply.cpp
runette: added in ef1fec4

mdal/frmts/mdal_ply.cpp
runette: added in ef1fec4

mdal/frmts/libplyxx/libplyxx.h
```diff
@@ -0,0 +1,226 @@
+#pragma once
```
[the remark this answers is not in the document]
runette: I was expecting and understand the concerns. I would make the following points :

- I personally would not count this as including the library as much as including code forked from that repository. I have extensively updated the code to integrate into MDAL, to add List Properties, to complete the set of data types that are supported and to fix a number bugs that did exist. So much so that, even if the source repository were actively managed this fork is now so far from the source that it can no longer take updates from the source. This is a self-contained driver that has to be viewed and maintained as stand-alone, that happens to include code (legally) forked from the original source - more of a template than a library.

- However, the basic structure using templates, function abstraction and maps, is sound and using that has saved considerable amounts of time putting the basic structure together. It is this that we are getting for free. Despite that, as part of the upgrades mentioned above, I have been through every line of the code, know what it does etc. There are no functions included that are not exclusively used for the process of reading or writing ASCII and BINARY PLY files and the same level of complexity would be required for a dedicated built-from-scratch driver, and the module design that this code brings makes the actual integration to the MDAL data model very simple, clear and flexible.

- I would note that the previous (read-only) driver was (at your insistence actually) copyright Runette Software. Legally (and ethically) this new driver is no different. It is code included into MDAL that is (from your point of view) 3rd party copyright. The only thing that has changed is the owner of the copyright. And the namespace - but that can be changed if it is a problem.

- Note that there are no "standard libraries" for reading PLY files. I did look. There are published example implementations - including Greg Turk's original - but actually, none of them is published under a licence that could be used as part of MDAL.

LICENSE
```diff
@@ -19,3 +19,43 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
PeterPetrik, line 61: remove.. and change line 3 to `Copyright (c) 2018-2021 MDAL Authors`
and add yourself and Simon to file https://github.com/lutraconsulting/MDAL/blob/master/AUTHORS (Simon with note about ply++)

mdal/frmts/libplyxx/libplyxx.h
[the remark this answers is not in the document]
PeterPetrik: OK to include it as it. But lets only add to the header of the files from libply++ that the file is `forked from the repo XYZ with MIT Licence Copyright Simon ...` Also I would not stick with the original names, lets rename them so it makes sense in context of MDAL.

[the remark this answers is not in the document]
nyalldawson: @runette 
> Note that there are no "standard libraries" for reading PLY files. I did look. There are published example implementations - including Greg Turk's original - but actually, none of them is published under a licence that could be used as part of MDAL.

There is https://github.com/phma/plytapus under a MIT license. (Not sure how mature it is though)

PeterPetrik:
Hi @runette any progress on this one? QGIS FF is half of Sept and in case you want to have it in the next release it would be good to have PR prepared for review at the beginning of September...

runette:
Lots of progress and it is almost ready - I will. send the next commit in
the next couple of days

mdal/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 # MDAL - Mesh Data Abstraction Library (MIT License)
 # Copyright (C) 2018 Peter Petrik (zilolv at gmail dot com)
 
+set(CMAKE_CXX_STANDARD 14)
```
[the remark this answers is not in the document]
runette: fixed in   974ee93

LICENSE
runette: Fixed in 974ee93

runette:
@PeterPetrik final code update committed (apart from that one style fix needed ) with tests updated. I am just working on the doc changes and will commit them soon.

This driver now :

- reads and saves meshes with any combination of Vertex / Face / Edge
- reads and saves both Vector and Scalar Vertex / Face / Edge datasets
- reads and saves Scalar (only) datasets on Volumes

runette:
@PeterPetrik Documentation update as well now.

docs/source/drivers/ply.rst
```diff
@@ -8,14 +8,15 @@ PLY -- ASCII Stanford Polygon Format
...
-MDAL supports reading of `PLY`_ format in ASCII format only. This format is used to store triangulated irregular networks (TINs) in ASCII format with arbitrary scalar data on the vertices, faces and/or edges.
+MDAL supports reading and writing of `PLY`_ format in ASCII and BINARY formats. This format is used to store triangulated irregular networks (TINs) in ASCII format with arbitrary scalar data on the vertices, faces and/or edges.
 
 Note that:
 
-- This driver does NOT support binary formats,
-- This driver does NOT support user-defined properties that are lists,
-- The `vertex-indices` property MUST be the first property defined on the face data.
-- This driver implements one additional feature without breaking the standard. If it finds a line in the header starting "comment crs " it will use the rest of the line as the string to set the mesh projection.
+- This driver can read and write Meshes containing any (constistent) combination of vertices, faces and edges and can access Scalar and Vector datasets on vertices, faces & edges and Scalar (only) datasets on volumes.
```
PeterPetrik, line 15: constistent -> consistent

mdal/CMakeLists.txt
```diff
@@ -37,6 +38,9 @@ SET(MDAL_HEADERS
   frmts/mdal_selafin.hpp
   frmts/mdal_esri_tin.hpp
   frmts/mdal_ply.hpp
+  3rdparty/libplyxx.h
+  3rdparty/libplyxx/libplyxx_internal.h
+  3rdparty/textio.h
```
PeterPetrik, line 43: can we introduce CMake flag "WITH_LIBPLYXX", so packagers can actually decided to build MDAL without PLY (in case of any issues with license, duplicated symbols used in libplyxx library or anything?)

mdal/frmts/mdal_ply.cpp
```diff
@@ -60,273 +69,236 @@ bool MDAL::DriverPly::canReadMesh( const std::string &uri )
...
-  }
 
-  /*
-  * The header is a format definition and a series of element definitions and/or comment lines
-  * cycle through these until end-header
-  */
-  do
+  //datastructures that will contain all of the datasets, categorised by vertex, face and edge datasets
+  std::vector<std::vector<double>> vertexDatasets; // contains the data
+  std::vector<std::pair<std::string, bool>> vProp2Ds; // contains the dataset name and a flag for scalr / vector
```
PeterPetrik, line 91: scalr - > scalar

mdal/frmts/mdal_ply.cpp
```diff
@@ -60,273 +69,236 @@ bool MDAL::DriverPly::canReadMesh( const std::string &uri )
...
-    MDAL::Log::error( MDAL_Status::Err_IncompatibleMesh, name(), meshFile + " Header is corrupt" );
-    return nullptr;
-  }
 
-  /*
-  * The header is a format definition and a series of element definitions and/or comment lines
-  * cycle through these until end-header
-  */
-  do
+  //datastructures that will contain all of the datasets, categorised by vertex, face and edge datasets
```
PeterPetrik, line 89: datastructures - > data structures

mdal/frmts/mdal_ply.cpp
```diff
@@ -60,273 +69,236 @@ bool MDAL::DriverPly::canReadMesh( const std::string &uri )
...
+  std::vector<std::vector<double>> vertexDatasets; // contains the data
+  std::vector<std::pair<std::string, bool>> vProp2Ds; // contains the dataset name and a flag for scalr / vector
+  std::vector<std::vector<double>> faceDatasets;
+  std::vector<std::pair<std::string, bool>> fProp2Ds;
+  std::vector<std::vector<double>> edgeDatasets;
+  std::vector<std::pair<std::string, bool>> eProp2Ds;
+  std::unordered_map<std::string, std::pair<std::vector<double>, std::vector<int>>> listProps; // contains the list datasets as vector of values and vector indices for each face into the vector of values
+
+  libply::File file( meshFile );
+  if ( MDAL::Log::getLastStatus() != MDAL_Status::None ) { return nullptr; }
```
PeterPetrik, line 99: you reseted the status at the beginning of the functions, it cannot be other than None?

mdal/frmts/mdal_ply.cpp
```diff
@@ -418,19 +462,19 @@ void MDAL::DriverPly::addDataset( MDAL::DatasetGroup *group, const std::vector<d
...
-    assert( values.size() == mesh->facesCount() );
+    assert( values.size() == mesh->facesCount()*mult );
     if ( mesh->facesCount() == 0 )
       return;
   }
 
   if ( group->dataLocation() == DataOnEdges )
   {
-    assert( values.size() == mesh->edgesCount() );
+    assert( values.size() == mesh->edgesCount()*mult );
```
PeterPetrik, line 477: maybe instead of assert move to the if() condition?

PeterPetrik (commented):
looks good!

mdal/frmts/mdal_ply.cpp
runette: Actually no - since libply::File has been adapted to raise errors through MDAL_Status - this line will pick up any file load errors.

docs/source/drivers/ply.rst
runette: Fixed in next commit

mdal/frmts/mdal_ply.cpp
runette: Fixed in next commit

mdal/frmts/mdal_ply.cpp
runette: Fixed in the next commit

mdal/frmts/mdal_ply.cpp
runette: Done in the next commit

PeterPetrik (approved):

PeterPetrik:
@runette can you please fix the tests and I will merge it afterwards?

mdal/CMakeLists.txt
runette: I made the flag "BUILD_PLY"

runette:
tests passing now

PeterPetrik:
thanks for the work Paul!

vcloarec:
I guess that means PLY file could be edited in QGIS. But the limitation will be that mesh editing in QGIS only support 2.5D.

runette:
It does mean that ply format meshes can be viewed in QGIS and edited in
Python - which does actually have some good implications for volumetric 3D
GIS processing

songololo:
@runette are any examples available showing how to edit meshes from python or `pyqgis`?

I'm currently trying to understand the `pyqgis` API in relation to mesh layers: I've so far figured out how to create the mesh but not how to add datasets or dataset groups.

Related question here: https://gis.stackexchange.com/questions/446044/qgsmeshlayer-how-to-add-dataset-groups-and-datasets-to-mesh

runette:
See the response to the related question :

The Python binding update to allow writing datasets has not been published
yet. Since you are now chasing this = I will push this through and publish
it. It was mainly waiting for me to work out how. to document it.

runette:
Slight correction to the above - I am forgetting my own projects :(. The bindings HAVE been updated to be write enabled. I just have not got around to updating the docs :(

PR #369, opened 2021-07-27
Add MemoryDataset3D

runette:
This PR addresses the issue of creating Volumetric datasets as discussed to some extent in [this conversation](https://lists.osgeo.org/pipermail/mdal-developer/2021-June/000034.html).

Currently - it is not possible to create 3D datasets through the C API because there is a hard-coded check to prevent it. This appears to be because there is no generic `MemoryDataset3D` to allow creating or writing these datasets.

Therefore - this PR :

- Adds `MemoryDataset3D`,
- Adds a new overload of `Driver::createDataset` to allow the creation of 3d datasets, and
- Adds a new call in the C API `MDAL_G_createDataset3D`

Together - this PR is essential to allow the writing of 3D datasets.

NOTE - since there are no drivers (yet) that can write 3D groups, this functionality cannot (yet) be tested through the C API (since it tests the capabilities). Interim tests have been added to the PLY driver tests using the C++ API, which will be updated when the write capable PLY driver is finalized. NOTE also that the interim tests (only - not the functionality) refuse to compile on Windows and I do not have the time to work the problem in what I hope is only a temporary test. The functionality of the PR does compile on Windows and has been tested. I just cannot get it to work in CI.

### DIFF
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -625,7 +625,7 @@ MDAL_EXPORT MDAL_DatasetH MDAL_G_addDataset(
  * \param time time for dataset (hours)
  * \param values For scalar data, the size must be volume count
  *               For vector data , the size must be volume count * 2 (x1, y1, x2, y2, ..., xN, yN)
- * \param verticalLevelCount Int Array holding the numver of vertical levels for each face.
+ * \param verticalLevelCount Int Array holding the number of vertical levels for each face.
  *               Size must be the face count
  * \param verticalExtrusion Double Array holding the vertical level values for the voxels
  *               Size must be Face count + Volume count
--- a/mdal/api/mdal.h
+++ b/mdal/api/mdal.h
@@ -611,6 +611,36 @@ MDAL_EXPORT MDAL_DatasetH MDAL_G_addDataset(
 );
 
 /**
+ * Adds empty (new) 3D dataset to the group
+ * This increases dataset group count MDAL_G_datasetCount() by 1
+ *
+ * The dataset is opened in edit mode.
+ * To persist dataset, call MDAL_G_closeEditMode() on parent group
+ *
+ * Minimum and maximum dataset values are automatically calculated
+ *
+ * Only for 3D datasets
+ *
+ * \param group parent group handle
+ * \param time time for dataset (hours)
+ * \param values For scalar data, the size must be volume count
+ *               For vector data , the size must be volume count * 2 (x1, y1, x2, y2, ..., xN, yN)
+ * \param verticalLevelCount Int Array holding the numver of vertical levels for each face.
+ *               Size must be the face count
+ * \param verticalExtrusion Double Array holding the vertical level values for the voxels
+ *               Size must be Face count + Volume count
+ * \returns empty pointer if not possible to create dataset (e.g. group opened in read mode), otherwise handle to new dataset
+ */
+
+MDAL_EXPORT MDAL_DatasetH MDAL_G_addDataset3D(
+  MDAL_DatasetGroupH group,
+  double time,
+  const double *values,
+  const int *verticalLevelCount,
+  const double *verticalExtrusions
+);
+
+/**
  * Returns whether dataset group is in edit mode
  */
 MDAL_EXPORT bool MDAL_G_isInEditMode( MDAL_DatasetGroupH group );
--- a/mdal/mdal.cpp
+++ b/mdal/mdal.cpp
@@ -884,12 +884,6 @@ MDAL_DatasetH MDAL_G_addDataset( MDAL_DatasetGroupH group, double time, const do
     return nullptr;
   }
 
-  if ( g->dataLocation() == MDAL_DataLocation::DataOnVolumes )
-  {
-    MDAL::Log::error( MDAL_Status::Err_MissingDriverCapability, "Dataset Group has data on 3D volumes" );
-    return nullptr;
-  }
-
   if ( active && g->dataLocation() != MDAL_DataLocation::DataOnVertices )
   {
     MDAL::Log::error( MDAL_Status::Err_IncompatibleDataset, "Active flag is only supported on datasets with data on vertices" );
--- a/mdal/mdal.cpp
+++ b/mdal/mdal.cpp
@@ -938,12 +938,6 @@ MDAL_DatasetH MDAL_G_addDataset3D( MDAL_DatasetGroupH group, double time, const 
     return nullptr;
   }
 
-  if ( !dr->hasWriteDatasetCapability( g->dataLocation() ) )
-  {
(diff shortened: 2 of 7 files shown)


mdal/api/mdal.h
```diff
@@ -610,6 +610,36 @@ MDAL_EXPORT MDAL_DatasetH MDAL_G_addDataset(
...
+ *
+ * Minimum and maximum dataset values are automatically calculated
+ *
+ * Only for 3D datasets
+ *
+ * \param group parent group handle
+ * \param time time for dataset (hours)
+ * \param values For scalar data, the size must be volume count
+ *               For vector data , the size must be volume count * 2 (x1, y1, x2, y2, ..., xN, yN)
+ * \param verticalLevelCount Int Array holding the numver of vertical levels for each face.
```
PeterPetrik, line 628: numver -> number

mdal/mdal.cpp
```diff
@@ -884,15 +884,15 @@ MDAL_DatasetH MDAL_G_addDataset( MDAL_DatasetGroupH group, double time, const do
...
     return nullptr;
   }
 
-  if ( active && g->dataLocation() != MDAL_DataLocation::DataOnVertices )
+  if ( g->dataLocation() == MDAL_DataLocation::DataOnVolumes )
   {
-    MDAL::Log::error( MDAL_Status::Err_IncompatibleDataset, "Active flag is only supported on datasets with data on vertices" );
+    MDAL::Log::error( MDAL_Status::Err_MissingDriverCapability, "Cannot save 3D dataset as a 2D dataset" );
     return nullptr;
   }
```
PeterPetrik, line 897: any reason this checks were switched?

mdal/mdal.cpp
```diff
@@ -909,6 +909,61 @@ MDAL_DatasetH MDAL_G_addDataset( MDAL_DatasetGroupH group, double time, const do
...
+{
+  if ( !group )
+  {
+    MDAL::Log::error( MDAL_Status::Err_IncompatibleDataset, "Dataset Group is not valid (null)" );
+    return nullptr;
+  }
+
+  if ( !values || !verticalLevelCount || !verticalExtrusions )
+  {
+    MDAL::Log::error( MDAL_Status::Err_InvalidData, "Passed pointer Values is not valid" );
```
PeterPetrik, line 922: is -> are

mdal/mdal.cpp
```diff
@@ -909,6 +909,61 @@ MDAL_DatasetH MDAL_G_addDataset( MDAL_DatasetGroupH group, double time, const do
...
+
+  const std::string driverName = g->driverName();
+  std::shared_ptr<MDAL::Driver> dr = MDAL::DriverManager::instance().driver( driverName );
+  if ( !dr )
+  {
+    MDAL::Log::error( MDAL_Status::Err_MissingDriver, "Driver name " + driverName + " saved in dataset group could not be found" );
+    return nullptr;
+  }
+
+  if ( !dr->hasWriteDatasetCapability( g->dataLocation() ) )
```
PeterPetrik, line 941: cannot we make new capabilit "write3DDatasetCapability" ? since I assume tons of drivers could write 2d, but not 3d

PeterPetrik (commented):
looks good to me. What was wrong with he windows tests? Looks a bit scary to push changes that are not tested on the main platform...

runette:
> looks good to me. What was wrong with the windows tests? Looks a bit scary to push changes that are not tested on the main platform...

The only error I cannot find a way to resolve is this 

```
test_ply.obj : error LNK2019: unresolved external symbol "public: class MDAL::RelativeTimestamp __cdecl MDAL::Dataset::
timestamp(void)const " (?timestamp@Dataset@MDAL@@QEBA?AVRelativeTimestamp@2@XZ) referenced in function "private: virtua
l void __cdecl Memory3D_ScalarMesh_Test::TestBody(void)" (?TestBody@Memory3D_ScalarMesh_Test@@EEAAXXZ) [C:\Users\runes\
Documents\GitHub\MDAL\build\tests\mdal_ply_test.vcxproj]
C:\Users\runes\Documents\GitHub\MDAL\build\tests\Release\mdal_ply_test.exe : fatal error LNK1120: 1 unresolved external
s [C:\Users\runes\Documents\GitHub\MDAL\build\tests\mdal_ply_test.vcxproj]
```

I am not sure why that is the symbol it cannot find. It successfully links to other MDAL::Dataset methods (although not all - it had a problem with volumesCount as well, but I can get that one through the C API). It is not a part of the code that I have changed in any way. In the end, since this problem WILL go away at the point when there is a driver with the DataOnVolumes capability and until then since the current tests are passing, this code does not affect any existing user.

As soon as this PR is accepted, I can push a new version of the PLY PR that does test this functionality - but obviously, there is a lot of work to do before that is accepted.

mdal/mdal.cpp
runette: I am not sure how that would differ from the current `WriteDatasetsOnVolumes`capability?https://github.com/lutraconsulting/MDAL/blob/0905f9fd3dc3bf44aa940bcd6929bd5da7b39c9b/mdal/frmts/mdal_driver.hpp#L23

mdal/mdal.cpp
runette: Laziness in the development process - I deleted the old and create a new one out of order - I will change it back for consistency

mdal/mdal.cpp
PeterPetrik: sure, sorry

mdal/api/mdal.h
runette: resolved in bc85fd85d3b09458e2d4ed04ed99644df0138a12

mdal/mdal.cpp
runette: resolved in bc85fd85d3b09458e2d4ed04ed99644df0138a12

mdal/mdal.cpp
runette: resolved in bc85fd85d3b09458e2d4ed04ed99644df0138a12

PeterPetrik:
@vcloarec do you have windows build machine setup? Could you please uncomment the `ifdef` for windows test and try to fix the build? Thanks!

runette:
I had a brain wave - or possibly a brainstorm - this morning.

I realized that the compile problem is an unintentional side effect of the fact that there is a test of write capability when adding a new dataset through the C API

HOWEVER - I also realized that this check is totally redundant since we also check if the DatasetGroup is in EditMode and it cannot be in EditMode if the driver does not have the write capability (again when working through the C API).

So - I have removed the redundant check in the C API addDataset3D method - this means that the tests use the C API to access the dataset, compile successfully on Windows and the test coverage is much high and redundant code has been removed.

The test is still - interim - using the C++ API to create the new DatasetGroup - since we cannot and should not remove the check for write capability in that method.

PeterPetrik (approved):
great!

Issue #361, opened 2021-06-21
Driver Capability data and documentation is inconsistent and incomplete


runette:
Working through adding the save function to the Python interface, it has become clear to me that the available documentation AND the underlying capability flags for Driver Capabilities are inconsistent and incomplete.

I created a small Python app to tabulate the underlying Capability flags as accessed through the C API and the results look like this where "Vertex Data" for instance means the write Capability for Vertex data since there is no flag for read capability

| Driver     | Description                  | Load Mesh   | Save Mesh   | Vertex Data   | Face Data   | Edge Date   | Volume Data   |
|------------|------------------------------|-------------|-------------|---------------|-------------|-------------|---------------|
| 2DM        | 2DM Mesh File                | True        | True        | False         | False       | False       | False         |
| XMS_TIN    | XMS Tin Mesh File            | True        | False       | False         | False       | False       | False         |
| SELAFIN    | Selafin File                 | True        | True        | True          | False       | False       | False         |
| ESRI_TIN   | Esri TIN                     | True        | False       | False         | False       | False       | False         |
| PLY        | Stanford PLY Ascii Mesh File | True        | False       | False         | False       | False       | False         |
| FLO2D      | Flo2D                        | True        | False       | False         | True        | False       | False         |
| HEC2D      | HEC-RAS 2D                   | True        | False       | False         | False       | False       | False         |
| TUFLOWFV   | TUFLOW FV                    | True        | False       | False         | False       | False       | False         |
| SWW        | AnuGA                        | True        | False       | False         | False       | False       | False         |
| Ugrid      | UGRID Results                | True        | True        | False         | False       | False       | False         |
| NETCDF     | GDAL NetCDF                  | True        | False       | False         | False       | False       | False         |
| GRIB       | GDAL Grib                    | True        | False       | False         | False       | False       | False         |
| ASCII_DAT  | DAT                          | False       | False       | True          | True        | True        | False         |
| BINARY_DAT | Binary DAT                   | False       | False       | True          | False       | False       | False         |
| XMDF       | TUFLOW XMDF                  | False       | False       | False         | False       | False       | False         |
| XDMF       | XDMF                         | False       | False       | False         | False       | False       | False         |

Some Issues:

[...]

PR #373, opened 2021-08-21
Add Edges and Edit Mode

### LINKED ISSUE #360: The C API (and the C++ API) do not seem to have the calls to addEdges
The C API has the calls MDAL_M_addVerices() and MDAL_M_addFaces() calls.

There does not seem to be an MDAL_M_addEdges() call and I cannot find the underlying C++ structure to do this.

This is despite at least one Driver advertising the ability to write Edge data.

Is there a fundamental reason behind this? 

If there is we should add it to the documentation.
If there is not then we should probably look at completing the API.

### LINKED ISSUE #365: The C API is not complete around Edit Modes
The handing of edit modes in the C API  needs to be improved:

- For a Mesh - it is created in Edit Mode but loaded not in Edit Mode. So the only way to edit a mesh is to create a new mesh. This IS workable and kind of makes sense because it avoids the complexities of data validation when editing, for instance, the vertices of a mesh with DataOnVertices etc. 

- The case for DataGroup is less clear. At the moment it is similar - created in Edit Mode but loaded read only. However, since read only prevents you from overwriting the complete dataset or adding a timeslice to the data (both perfectly reasonable actions) I think that this is not a sensible choice. The API should have the option to open Edit Mode as well as close Edit Mode.

I will submit a separate PR with proposed changes.

runette:
fixes #360 

This should be the last of updates needed for the next Python version.

- Adds the C and C++ methods to add edges to an editable mesh
- Updates and extends the API tests to include the edge API


mdal/api/mdal.h
```diff
@@ -397,6 +413,8 @@ MDAL_EXPORT int MDAL_M_datasetGroupCount( MDAL_MeshH mesh );
 
 /**
  * Returns dataset group handle
+ *
+ * Since 0.9.0 - if the driver has Write Capability for this DatasetGroup, it is opened in edit mode
```
PeterPetrik, line 417: I strongly suggest to not change the behaviour of the exisisting functions. they are used all over the place in qgis and we would need to make sure to close edit mode? Also if you want to only read the dataset you cannot do it. Can you create a new function to start edit mode of existing mesh dataset group instead?

mdal/api/mdal.h
```diff
@@ -647,6 +665,8 @@ MDAL_EXPORT MDAL_DatasetH MDAL_G_addDataset(
  * \param verticalExtrusion Double Array holding the vertical level values for the voxels
  *               Size must be Face count + Volume count
  * \returns empty pointer if not possible to create dataset (e.g. group opened in read mode), otherwise handle to new dataset
+ *
+ * \since MDAL 0.9.0
```
PeterPetrik, line 669: this is added in this version?

tests/test_api.cpp
```diff
@@ -242,6 +242,91 @@ TEST( ApiTest, VerticesApi )
...
+                  );
+
+    EXPECT_TRUE( compareVectors( refStart, start ) );
+    EXPECT_TRUE( compareVectors( refStop, stop ) );
+  }
+  MDAL_CloseMesh( m );
+
+  // Some wrong calls tests
+  EXPECT_EQ( MDAL_M_edgeIterator( nullptr ), nullptr );
+  EXPECT_EQ( MDAL_EI_next( nullptr, 0, nullptr, nullptr ), 0 );
```
PeterPetrik, line 327: should be separate test case and not mixed with regular tests

tests/test_api.cpp
```diff
@@ -242,6 +242,91 @@ TEST( ApiTest, VerticesApi )
   EXPECT_EQ( MDAL_FI_next( nullptr, 0, nullptr, 0, nullptr ), 0 );
 }
 
+void _populateEdges( MDAL_MeshH m, std::vector<int> &start, std::vector<int> &stop, size_t itemsLen )
```
PeterPetrik, line 245: static?

mdal/mdal_data_model.hpp
```diff
@@ -279,6 +279,8 @@ namespace MDAL
 
       virtual void addVertices( size_t vertexCount, double *coordinates );
       virtual void addFaces( size_t faceCount, size_t driverMaxVerticesPerFace, int *faceSizes, int *vertexIndices );
+      virtual void addEdges( size_t edgeCount, int *startVertexIndices, int *endVertexIndices );
```
PeterPetrik, line 282: wondering if we shouldn't add new capability in https://github.com/lutraconsulting/MDAL/blob/5e8223ed283c15815089bcf7e5558e4b5660ac88/mdal/frmts/mdal_driver.hpp#L15 that driver is able to modify/write mesh frame (add faces/edges/vertices) ..? so at least we can know how we stand with driver support

PeterPetrik (commented):
Thanks, I will have more time later this week to review your work!

mdal/api/mdal.h
runette: Ok - I removed these changes from this PR and will add something else later

tests/test_api.cpp
runette: Not totally sure which part you are referring to and what you mean.

mdal/mdal_data_model.hpp
runette: I do agree with you - see #361 - but I don't think that should be fixed in this PR and/or only for edges (which would create inconsistency) I suggest a separate dedicated PR to update the capabilities across all entities.

mdal/api/mdal.h
runette: I think that is a copy/paste error. I will correct it.

runette: I have to correct myself - that is in the comments which are new in version 0.9.0 but in a previous PR. I just noticed that the version tag was not added in that PR

tests/test_api.cpp
PeterPetrik: just that I would not mix in on test function both tests that check that the api is working well with invalid inputs e.g. `EXPECT_EQ( MDAL_M_edgeIterator( nullptr ), nullptr );` and a normal test case where you test the api works for regular use case.

PeterPetrik: but it is minor, do not worry

mdal/mdal_data_model.hpp
PeterPetrik: ok

mdal/api/mdal.h
PeterPetrik: can you also please change the api doc for this function accordingly?

PeterPetrik: @runette please change the api doc for this function accordingly? and i will merge afterwards

runette: API Doc corrected - good catch

PeterPetrik (approved):

PR #374, opened 2021-09-06
fix 2DM write when face size > 4

vcloarec:
Fixs to allow 2D save faces that have more than 4 vertices (up to 6 for reading).
Also allows face with 5 vertices, not such face in 2DM specification, but added here to avoid non saved mesh 
(for example created with mesh editing with QGIS)


PeterPetrik:
I would not like to add something that is not in the specification or at least by any solver as a custom extension. Why QGIS cannot gracefully tell user to choose different driver?

PR #384, opened 2021-10-14
handle special char for windows

vcloarec:
Only for drivers where open files is handle directly in MDAL. For drivers that depend on external library, that will depends if these libraries support special char...


nyalldawson:
@vcloarec Nope, no special characters (aside from backslashes). Could it potentially be a path length issue?

vcloarec:
> Could it potentially be a path length issue?

I have to dig a bit in Windows stuff... The best will be to have the string path...

Issue #381, opened 2021-10-06
[FLO-2D] Crash on data loading 


PeterPetrik:
https://github.com/lutraconsulting/MDAL/blob/4d3511f04ff4daf5be484e0679f9db8942301710/mdal/frmts/mdal_flo2d.cpp#L862

PR #385, opened 2021-10-16
fix flo2d rounding error

vcloarec:
alternative fix to #383

Issue came from a rounding error when calculating the the width/height of the mesh:
 `size_t(1542.9999 + 1 ) -> 1543` while the real width is 1544
with 
`size_t(1542.9999 + 1.5 ) -> 1544` -> ok in any cases

I think it is better to keep the old approach to populate the vertices. Indeed the new approach of #383 is quite slower.
Some tests test with files that leads to  #381, the time just to populate vertices (release build):
- #383: 150 ms
- old approach: 25 ms


vcloarec:
> so the memory usage is probably higher, right?

not sure, here the extra memory usage is `verticesCount x size_of(size_t) `, 
with the std::map aproach, the extra memory usage in #383 is `verticesCount x size_of(std::pair<size_t,size_t>) `, no?

PR #393, opened 2021-12-02
Add extensions to SELAFIN driver

nicogodet:
This PR will allow QGIS to open more SELAFIN file with different extension.
As a single TELEMAC model is multi file, these added extensions are quite commonly used to differentiate files inside the model.

- `.ser` : SERAFIN, the other name of SELAFIN
- `.geo` : used for the geometry file (usually no dataset, juste the mesh)
- `.res` : the result file

Extension does not matter for TELEMAC.

### DIFF
--- a/mdal/frmts/mdal_selafin.cpp
+++ b/mdal/frmts/mdal_selafin.cpp
@@ -615,7 +615,7 @@ void MDAL::SelafinFile::ignoreArrayLength( )
 MDAL::DriverSelafin::DriverSelafin():
   Driver( "SELAFIN",
           "Selafin File",
-          "*.slf",
+          "*.slf;;*.ser;;*.geo;;*.res",
           Capability::ReadMesh | Capability::SaveMesh | Capability::WriteDatasetsOnVertices | Capability::ReadDatasets
         )
 {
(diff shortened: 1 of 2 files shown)


nicogodet:
Not sure how this PR can affect that much.
Are all these test failure related to my changes ?

mdal/frmts/mdal_selafin.cpp
```diff
@@ -615,7 +615,7 @@ void MDAL::SelafinFile::ignoreArrayLength( )
 MDAL::DriverSelafin::DriverSelafin():
   Driver( "SELAFIN",
           "Selafin File",
-          "*.slf",
+          "*.slf;;*.ser;;*.geo;;*.res",
```
PeterPetrik, line 618: I understand *.ser, but does "geo" and "res" file has exactly the same internal file structure as slf file? If not, than the driver will not be able to open the files

nicogodet: @PeterPetrik Yes. The internal structure is exactly the same.

It's the header inside the binary file that determine if the file is a SELAFIN file, not the extension.

PeterPetrik (approved):

PR #400, opened 2022-01-25
H2i format

vcloarec:
H2i mesh structure is a quad tree structure. The structure is defined by nodes a the center of quad and by links that, at each side of quads, represents the interface between quad.
  
Files of H2i format are:
- a json file with:
        - some information about the mesh (mesh name, crs, reference time and time steps)
        - relative path of the files containing the nodes definition and the links definition
        - relative pathto the file containing the quantity dataset groups.
- text file containing the nodes information
- text file containing the links information
- text file containing the time step
- one binary file per dataset group, dataset group can be relied to nodes or links.
  
Dataset groups on nodes have scalar values (water depth, water level) that are interpreted by MDAL as scalar value on faces.

Dataset groups on links have also scalar values but related to quantities that are vectors. Theses values as interpreted by MDAL as vector value on faces using direction of the link and converting value like this:
for each face (node), X or Y vector value is average of value of link in X or Y direction weighted by the side lenght of the node on the other side of the link.

h2i

For velocity, the interpolation is done with this formula:
formula_H2i_V

where
 - Si vertical side sizes,
 - u  values on horizontal link related to Si
 - Sj horizontal side sizes
 - v values on vertical link related to Sj
 - V, the vector value on face for MDAL

for a quad cell with a size S, sum(Si)=2S and sum(S) =2S, so:
formula_H2i_V_simpl

For the particular case of discharge, H2i format store this quantity in links as full discharge (m3/s), that is not compatible with a 2D planar representation.
So MDAL converts discharge dataset group as "unit discharge"(m2/s), dividing the full discharge value per the side of faces to retrieve the unit discharge:

formula_H2i_qQ
where 
- Q is full flow values on links
- q, the unit flow in X or Y direction
- S, the size of the side
 
Then, to interpolate on the face, with the same principle as velocity:

formula_H2i_q


mdal/frmts/mdal_h2i.cpp
```diff
@@ -0,0 +1,575 @@
...
+    const std::vector<std::string> lineElements = split( line, ' ' );
+    if ( lineElements.size() != 7 ) throw MDAL::Error( MDAL_Status::Err_UnknownFormat, "File format not recognized: " + filePath );
+
+    CellH2i cell;
+    cell.x = toDouble( lineElements.at( 1 ) );
+    cell.y = toDouble( lineElements.at( 2 ) );
+    cell.size = toDouble( lineElements.at( 4 ) );
+    cell.z = ( toDouble( lineElements.at( 5 ) ) + toDouble( lineElements.at( 6 ) ) ) / 2;
+
+    cell.neighborsCellCountperSide = std::vector<int>( 4, 0 );
```
PeterPetrik, line 371: typo: neighborsCellCountPerSide

tests/CMakeLists.txt
```diff
@@ -126,7 +127,7 @@ IF(BUILD_EXTERNAL_DRIVERS)
     test_dynamic_driver.cpp
   )
 
-  IF(EXTERNAL_DRIVER_DHI_DFS)
+  IF(EXTERNAL_DRIVER_DHI_DFSU)
```
PeterPetrik, line 130: this is leftover/bug?

PeterPetrik (commented):
Looks good!

Would it be possible to add new driver description to https://github.com/lutraconsulting/MDAL/tree/master/docs/source/drivers and https://github.com/lutraconsulting/MDAL/blob/master/docs/source/drivers/index.rst

and also add it to the list on https://github.com/lutraconsulting/MDAL/blob/master/README.md#supported-formats

PeterPetrik:
please do nott use force-push once the PR is created, only just before mergin to master. with force-push I am unable to see the changes between the original PR I reviewed, so I need to go through all of it potentially. Thanks

PeterPetrik:
Would it be possible to add it to the docs/README?

PeterPetrik (approved):

PR #401, opened 2022-01-25
ver++ 0.9.2

vcloarec:
ver++ MDAL 0.9.2 a bit late...


runette:
Is there anyway of backporting this to the tarball - it explains why I was having trouble updating one of my subsidiary packages that does a version check on install!

PR #403, opened 2022-02-03
Fix format check for HEC-RAS.

freekvw:
I have a file with version "HEC-RAS 6.0.0 May 2021" that has file type "HEC-RAS Results", but uses the new way of storing flow area names.  This PR changes the format check to look for a "Names" dataset instead of checking the file type string.

### DIFF
--- a/mdal/frmts/mdal_hec2d.cpp
+++ b/mdal/frmts/mdal_hec2d.cpp
@@ -44,6 +44,11 @@ static HdfDataset openHdfDataset( const HdfGroup &hdfGroup, const std::string &n
   return dsFileType;
 }
 
+static bool existsHdfDataset( const HdfGroup &hdfGroup, const std::string &name )
+{
+  return hdfGroup.dataset( name ).isValid();
+}
+
 static std::string openHdfAttribute( const HdfFile &hdfFile, const std::string &name )
 {
   HdfAttribute attr = hdfFile.attribute( name );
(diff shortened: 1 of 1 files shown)


PeterPetrik:
Hi @freekvw thank you for your contribution, before accepting the PR, we would need to add a small example test file (<3mb) to the HEC-RAS folder with the file you want to support and also add a test to this new file in the hec-ras automatic test suite.

mdal/frmts/mdal_hec2d.cpp
```diff
@@ -761,13 +766,11 @@ std::unique_ptr<MDAL::Mesh> MDAL::DriverHec2D::load( const std::string &resultsF
...
     HdfFile hdfFile = openHdfFile( mFileName );
 
-    // Verify it is correct file
-    std::string fileType = openHdfAttribute( hdfFile, "File Type" );
-    bool oldFormat = canReadOldFormat( fileType );
-
     HdfGroup gGeom = openHdfGroup( hdfFile, "Geometry" );
     HdfGroup gGeom2DFlowAreas = openHdfGroup( gGeom, "2D Flow Areas" );
 
+    bool oldFormat( existsHdfDataset( gGeom2DFlowAreas, "Names" ) );
```
PeterPetrik, line 772: are you sure that this format you want to support is exactly same as "old format" in all other spec? if it is different format that 2 we currently support ("old" and "new"), maybe there is need to add 3rd format spec .... ?

PeterPetrik:
Hi @freekvw would you like to join our Slack chat to discuss your PR in more details? If you send me mail to peter.petrik@lutraconsulting.co.uk I can add you there?

freekvw:
Hi Peter,
I believe the current test for "File Type" is not the right test to distinguish the old and new format.  Maybe the File Type was different for a beta version of HEC-RAS 6, but on 6.0.0 and 6.1.0, the File Type is still "HEC-RAS Results".
I managed to create a small example with HEC-RAS 6.1.0: [Muncie2DOnly_SI.zip](https://github.com/lutraconsulting/MDAL/files/8031210/Muncie2DOnly_SI.zip)

This PR simply proposes to change the test in a way that makes it work in all known cases.  Alternatively one could maybe check the "File Version" string.

PeterPetrik:
Hi @vcloarec can you review this PR please?

vcloarec:
thanks @freekvw to pointing to this.
It seems HEC-RAS change a bit its format each version, and MDAL becomes confused about recognizing format. 
If I well understand, the file type was "`HEC-RAS Results`" before 5.05, became "`HEC-RAS Geometry`" with 5.05 and came back to "`HEC-RAS Results`" after 6.0 beta...
And for the "Names" dataset, not clear...
With all this changes, the approach in MDAL became confusing, so I purpose the approach https://github.com/lutraconsulting/MDAL/pull/404 to fix this issue and clean an bit the approach.
I added your file in the unit tests and all seems ok.

PeterPetrik:
@freekvw thanks, let me close this PR if you agree, as the @vcloarec PR has also test case and it works for all the formats.

PR #404, opened 2022-02-10
fix Hec 6.1, cleaner approach to read hec file

vcloarec:


mdal/frmts/mdal_hec2d.cpp
```diff
@@ -761,18 +760,30 @@ std::unique_ptr<MDAL::Mesh> MDAL::DriverHec2D::load( const std::string &resultsF
...
 
     std::vector<std::string> flowAreaNames;
-    if ( oldFormat )
-      flowAreaNames = read2DFlowAreasNamesOld( gGeom2DFlowAreas );
-    else
-      flowAreaNames = read2DFlowAreasNames505( gGeom2DFlowAreas );
+
+    try
+    {
+      flowAreaNames = read2DFlowAreasNamesFromNameDataset( gGeom2DFlowAreas );
```
PeterPetrik, line 770: can we change this read2DFlowAreasNamesFromNameDataset() and read2DFlowAreasNamesFromAttributesDataset() to just return empty flowAreaNames on error and not throw? 

it is generally not advisable to use try-catch for expected code-flow and not error situation.

PeterPetrik (commented):
this looks ok

mdal/frmts/mdal_hec2d.cpp
vcloarec: done

PeterPetrik (approved):

vcloarec:
windows test is now fix and lives in 2022.

PR #405, opened 2022-02-16
Revert "H2i format (#400)"

vcloarec:
This reverts commit b7e124fd910f2bcd4b8e5f67abfd0d8ddf4976c7.

As this format is very experimental for now and also need to be rewritten because of big changes, it is preferable to remove it from master before release 0.9.3 and allow secure backport of this version to QGIS 3.22.
Then we could continue dev on master (0.9.90) and we could keep a release 0.9.X branch for QGIS 3.22.


PeterPetrik:
Can you:
- create branch release_0_9 and create this PR (revert) to this branch. Also bump the version to 0.9.4 on the branch. So the revert madness will not go through main.
- bump the version on master to 1.0.90

PR #410, opened 2022-03-02
bump to 0.9.90

vcloarec:
some doc and indentation fixes, other clean and version++


wonder-sk (approved):

wonder-sk:
Maybe worth mentioning in docs that the format spec is here: https://github.com/d2hydro/H2i_code_factory/blob/main/docs/data_format.md

Issue #412, opened 2022-03-12
failing to build 0.9.4 on linux (external_drivers/minimal_example/CMakeFiles/mdal_dummy_driver.dir/mdalDummyExternalDriver.cpp.o)


okanisis:
### Version Info:

- gdal 3.4.1
- netcdf 4.8.1
- hdf5 1.12.1
- libxml2 2.9.13
- sqlite 3.38.0

### cmake:
```
    cmake -B build -S "${pkgname}-${pkgver}" \
        -DCMAKE_BUILD_TYPE='None' \
        -DCMAKE_INSTALL_PREFIX='/usr' \
        -DENABLE_TESTS=ON \
        -DSQLITE3_LIBRARY=/usr/lib/libsqlite3.so \
        -Wno-dev
```

### Error:
Getting the following error trying to build on Arch Linux 64-bit with cmake.

```
[ 99%] Building CXX object external_drivers/minimal_example/CMakeFiles/mdal_dummy_driver.dir/mdalDummyExternalDriver.cpp.o
/build/mdal/src/mdal-0.9.4/external_drivers/minimal_example/mdalDummyExternalDriver.cpp: In function ‘void MDAL_DRIVER_M_extent(int, double*, double*, double*, double*)’:
/build/mdal/src/mdal-0.9.4/external_drivers/minimal_example/mdalDummyExternalDriver.cpp:305:16: error: ‘numeric_limits’ is not a member of ‘std’
  305 |   *xMin = std::numeric_limits<double>::quiet_NaN();
      |                ^~~~~~~~~~~~~~
/build/mdal/src/mdal-0.9.4/external_drivers/minimal_example/mdalDummyExternalDriver.cpp:305:31: error: expected primary-expression before ‘double’
  305 |   *xMin = std::numeric_limits<double>::quiet_NaN();
      |                               ^~~~~~
/build/mdal/src/mdal-0.9.4/external_drivers/minimal_example/mdalDummyExternalDriver.cpp:306:16: error: ‘numeric_limits’ is not a member of ‘std’
  306 |   *xMax = std::numeric_limits<double>::quiet_NaN();
      |                ^~~~~~~~~~~~~~
/build/mdal/src/mdal-0.9.4/external_drivers/minimal_example/mdalDummyExternalDriver.cpp:306:31: error: expected primary-expression before ‘double’
  306 |   *xMax = std::numeric_limits<double>::quiet_NaN();
      |                               ^~~~~~
[...]

PR #413, opened 2022-03-13
potential fix for #412

okanisis:
my first pull request so understand if it's not up to par


external_drivers/minimal_example/mdalDummyExternalDriver.cpp
```diff
@@ -13,7 +13,7 @@
 
 #include "mdal_external_driver.h"
 
-#include "limits.h"
+#include <limits>
```
vcloarec, line 16: only a form suggestion, can you just move this #include to line 13? to have std include together.

vcloarec (approved):

PR #418, opened 2022-07-11
Take over the validity check for numdata in MDAL::DriverBinaryDat::lo…

wilhelmje:
…ad from MDAL::DriverAsciiDat::loadOldFormat and MDAL::DriverAsciiDat::loadNewFormat


wilhelmje:
Sorry.

PR #419, opened 2022-08-02
Write ugrid dataset group

vcloarec:
This PR add the possibility to persist dataset groups for 2D mesh UGRID format

### DIFF
--- a/mdal/frmts/mdal_ugrid.cpp
+++ b/mdal/frmts/mdal_ugrid.cpp
@@ -176,7 +176,7 @@ MDAL::CFDimensions MDAL::DriverUgrid::populateDimensions( )
   return dims;
 }
 
-void MDAL::DriverUgrid::populate1DMeshDimensions( MDAL::CFDimensions &dims )
+void MDAL::DriverUgrid::populate1DMeshDimensions( MDAL::CFDimensions &dims ) const
 {
   /* Parse number of edges ( dimension ) from mesh */
   std::string edgeConnectivityVariableName = mNcFile->getAttrStr( mMeshName, "edge_node_connectivity" );
@@ -195,7 +195,7 @@ void MDAL::DriverUgrid::populate1DMeshDimensions( MDAL::CFDimensions &dims )
   dims.setDimension( CFDimensions::Edge, edgesCount, edgesCountId );
 }
 
-void MDAL::DriverUgrid::populate2DMeshDimensions( MDAL::CFDimensions &dims, int &ncid )
+void MDAL::DriverUgrid::populate2DMeshDimensions( MDAL::CFDimensions &dims, int &ncid ) const
 {
   // face dimension location is retrieved from the face_node_connectivity variable
   // if face_dimension is defined as attribute, the dimension at this location help to desambiguate vertex per faces and number of faces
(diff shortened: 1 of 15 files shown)


mdal/frmts/mdal_ugrid.cpp
```diff
@@ -695,65 +702,275 @@ void MDAL::DriverUgrid::save( const std::string &fileName, const std::string &,
...
+  try
+  {
+    mNcFile->openFile( fileName, true );
+  }
+  catch ( MDAL::Error &err )
+  {
+    err.setDriver( name() );
+    MDAL::Log::error( err );
+    return true;
+  }
```
PeterPetrik, line 732: I would normally assume that you open file once in the persist function, add all groups/variables and then close a file. Why you need to open/close files for each dataset group? is it consistent with other drivers?

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -782,10 +999,6 @@ void MDAL::DriverUgrid::writeVariables( MDAL::Mesh *mesh )
     }
   }
 
-  // Time array
-  int timeId = mNcFile->defineVar( "time", NC_DOUBLE, 1, &dimTimeId );
-  mNcFile->putAttrStr( timeId, "units", "hours since 2000-01-01 00:00:00" );
-
```
PeterPetrik, line 788: why this is removed?

mdal/frmts/mdal_ugrid.cpp
```diff
@@ -878,12 +1091,6 @@ void MDAL::DriverUgrid::writeVariables( MDAL::Mesh *mesh )
       faceIndex += facesRead;
     }
   }
-
-  // Time values (not implemented)
-  mNcFile->putDataDouble( timeId, 0, 0.0 );
-
-  // Turning on define mode
-  nc_redef( mNcFile->handle() );
```
PeterPetrik, line 886: why this is removed?

tests/test_ugrid.cpp
```diff
@@ -814,6 +814,306 @@ TEST( MeshUgridTest, flow_3d )
...
+  MDAL_CloseMesh( m );
+
+  m = MDAL_LoadMesh( tmpFile.c_str() );
+  ASSERT_NE( m, nullptr );
+  s = MDAL_LastStatus();
+  EXPECT_EQ( MDAL_Status::None, s );
+
+  groupCount = MDAL_M_datasetGroupCount( m );
+  ASSERT_EQ( 14, groupCount );
+
```
PeterPetrik, line 966: would be good to check at least one scalar, one vector value on various data locations

tests/test_ugrid.cpp
```diff
@@ -814,6 +814,306 @@ TEST( MeshUgridTest, flow_3d )
...
+  EXPECT_TRUE( success );
+
+  createNewDatasetGroupOnNewFile( m, tmpFile3, refTime, timeSteps, "new group vector faces", MDAL_DataLocation::DataOnFaces, faceCount, false, 1.23, success );
+  EXPECT_TRUE( success );
+
+  createNewDatasetGroupOnNewFile( m, tmpFile4, refTime, timeSteps, "new group vector vertices", MDAL_DataLocation::DataOnVertices, vertexCount, false, 4.56, success );
+  EXPECT_TRUE( success );
+
+  MDAL_CloseMesh( m );
+
```
PeterPetrik, line 1110: would be good to load the written file and check few values/variables?

mdal/frmts/mdal_ugrid.cpp
vcloarec: Same logic for other drivers. Persisting dataset group concerns only one dataset group.

mdal/frmts/mdal_ugrid.cpp
vcloarec: This method is called only for writing mesh frame. In that case, there is no time, and I think UGRID supports no time. Further more, settings a dummy time here leads to not be able to check if the time is the same when writing new dataset group.
The logic is:
- if there is already times in the file, MDAL checks if the times (ref and time steps) are the same, error if not
- if there is no times, adds time depending of the times of the new dataset group.

mdal/frmts/mdal_ugrid.cpp
PeterPetrik: ok cool

mdal/frmts/mdal_ugrid.cpp
vcloarec: for times, same reason as above.

For define mode, I think it is an old left over. Define mode has to be set before adding new variables, not at the end.

tests/test_ugrid.cpp
vcloarec: dataset values are checked in `createNewDatasetGroupOnExistingFile()`

tests/test_ugrid.cpp
vcloarec: dataset values are checked in `createNewDatasetGroupOnNewFile()`

Issue #348, opened 2021-03-31
UGRID reader does not support time coordinates as integers


JoerivanEngelen:
Hi,

We encountered an issue in the UGRID reader when the time coordinates are encoded as integers instead of doubles.
It seems like the UGRID reader expects time to be encoded as doubles. 

I have attached two small examples, one with time encoded as integers, one with doubles. For the version with integers, running `mdalinfo` throws a `ERROR: Status 3: Driver: Ugrid: Could not read double array`. MDAL then falls back to interpreting the file as a regular NetCDF. For the version with time as doubles, there are no problems.

[mdal_time_dtype.zip](https://github.com/lutraconsulting/MDAL/files/6237046/mdal_time_dtype.zip)

To my understanding, the UGRID conventions have no specifications for data types of time, so the CF conventions apply here. 
For which I can only find [here](http://cfconventions.org/cf-conventions/cf-conventions.html#_data_types)

> Data variables must be one of the following data types: string, char, byte, unsigned byte, short, unsigned short, int, unsigned int, int64, unsigned int64, float or real, and double

So I guess assuming the time coordinate is specified as doubles will result in issues for the naive user. For example, at our department, we use the python package [xarray](http://xarray.pydata.org/en/stable) a lot, which encodes CF time coordinates by default as integers.

visr:
The same `ERROR: Status 3: Driver: Ugrid: Could not read double array` also pops up in files without time coordinates. For instance, if I have a vector of type `int` with one element for every node.

The error disappears when converting the vector to floats. That suggests to me this is the same issue, and the problem isn't time specific.

vcloarec:
#420 seems not fix it.

PR #420, opened 2022-08-04
Fix bug #348

NEDJIMAbelgacem:
fixes bug #348


mdal/frmts/mdal_netcdf.cpp
```diff
@@ -160,6 +196,42 @@ std::vector<double> NetCDFFile::readDoubleArr( int arr_id,
   {
     if ( nc_get_vars_double( mNcid, arr_id, startp.data(), countp.data(), stridep.data(), arr_val.data() ) != NC_NOERR ) throw MDAL::Error( MDAL_Status::Err_UnknownFormat, "Could not read double array" );
   }
+  else if ( typep == NC_INT )
+  {
+    std::vector<int> arr_val_int( count_dim1 * count_dim2 );
+    if ( nc_get_vars_int( mNcid, arr_id, startp.data(), countp.data(), stridep.data(), arr_val_int.data() ) != NC_NOERR ) throw MDAL::Error( MDAL_Status::Err_UnknownFormat, "Could not read integer array" );
```
PeterPetrik, line 202: all errors should say "Could not read double array"

tests/test_ugrid.cpp
```diff
@@ -814,6 +814,30 @@ TEST( MeshUgridTest, flow_3d )
...
+  ASSERT_NE( m, nullptr );
+  MDAL_Status s = MDAL_LastStatus();
+  EXPECT_EQ( MDAL_Status::None, s );
+
+  int v_count = MDAL_M_vertexCount( m );
+  EXPECT_EQ( v_count, 6 );
+
+  int f_count = MDAL_M_faceCount( m );
+  EXPECT_EQ( f_count, 2 );
+
```
PeterPetrik, line 836: also test a value that was stored as int in the file and check it is correct double value!

tests/test_ugrid.cpp
```diff
@@ -814,6 +814,30 @@ TEST( MeshUgridTest, flow_3d )
   MDAL_CloseMesh( mesh );
 }
 
+TEST( MeshUgridTest, IntegerTimeMesh )
+{
+  std::string path = test_file( "/ugrid/time_integer/simple_time_integer.nc" );
```
PeterPetrik, line 819: would it be possible add to this file NC_UINT, NC_INT64, NC_UINT64 extra datasets and check that the values read from them are also correct?

tests/test_ugrid.cpp
```diff
@@ -814,6 +814,30 @@ TEST( MeshUgridTest, flow_3d )
   MDAL_CloseMesh( mesh );
 }
 
+TEST( MeshUgridTest, IntegerTimeMesh )
```
PeterPetrik, line 817: test coverage % dropped, see https://coveralls.io/builds/51423520/source?filename=mdal%2Ffrmts%2Fmdal_netcdf.cpp 

would it be possible to add some testcases that covers the code you have written?

PR #423, opened 2022-08-15
Fix mesh file staying open in cf driver

NEDJIMAbelgacem:
After loading a netcdf file it stays open because the driver will keep it open after calling DriverCF::canReadMesh.
fixes https://github.com/qgis/QGIS/issues/49215

### DIFF
--- a/mdal/frmts/mdal_cf.cpp
+++ b/mdal/frmts/mdal_cf.cpp
@@ -468,6 +468,7 @@ bool MDAL::DriverCF::canReadMesh( const std::string &uri )
     mNcFile.reset( new NetCDFFile );
     mNcFile->openFile( uri );
     populateDimensions( );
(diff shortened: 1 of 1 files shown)


vcloarec:
maybe we should do the same when leaving `MDAL::DriverCF::load()`?

PeterPetrik:
Is this deprecated by https://github.com/lutraconsulting/MDAL/pull/424 and we can close it? or we merge this one?

vcloarec:
> Is this deprecated by #424 and we can close it? or we merge this one?

No this one is not deprecated, same issue but not the same origin. 
And here, maybe it is better to add a test like in #424.

PeterPetrik:
Belgacem will not work on this one anymore, @vcloarec can you please finish it?

PeterPetrik (approved):

PR #424, opened 2022-09-02
Fix xmdf lock

vcloarec:
For Xmdf format, when the mesh is closed (for example, removed from QGIS), the files is still locked as it is still open, and user is unable to rename or remove the file until QGIS is closed.

Indeed, the mesh is not closed properly because the file's handle is closed before the dataset's handle are closed, leading to keep the file open.

To close the mesh properly it necessary to close the file's handle once every other Hdf5 handles are closed. To achieve this, a "shared file's handle" is embeded in each HdFGroup and each HdfDataset. This member is declared before the group/dataset handle to be sure the shared handle is destroy once the group/dataset handle is effectively closed.
The destruction of the last shared handle leads to close the file.

To improve the encapsulation around HdfFile, some refactorization are also done.

### DIFF
--- a/mdal/frmts/mdal_hdf5.hpp
+++ b/mdal/frmts/mdal_hdf5.hpp
@@ -139,7 +139,7 @@ class HdfGroup
     HdfGroup( std::shared_ptr<Handle> handle, HdfFile::SharedHandle file );
 
   private:
-    HdfFile::SharedHandle mFile; //must be the last destroyed
+    HdfFile::SharedHandle mFile; //must be declared before "std::shared_ptr<Handle> d" to be sure it will be the last destroyed
 
   protected:
     std::shared_ptr<Handle> d;
@@ -300,7 +300,7 @@ class HdfDataset
     HdfDataset( HdfFile::SharedHandle file, const std::string &path );
 
   private:
-    HdfFile::SharedHandle mFile; //must be the last destroyed
+    HdfFile::SharedHandle mFile; //must be declared before "std::shared_ptr<Handle> d" to be sure it will be the last destroyed
 
   protected:
     std::shared_ptr<Handle> d;
--- a/mdal/frmts/mdal_hdf5.hpp
+++ b/mdal/frmts/mdal_hdf5.hpp
@@ -66,7 +66,7 @@ class HdfFile
       Create
     };
     typedef HdfH<H5I_FILE> Handle;
-    typedef std::shared_ptr<HdfH<H5I_FILE>> SharedHandle;
+    typedef std::shared_ptr<Handle> SharedHandle;
 
     HdfFile( const std::string &path, HdfFile::Mode mode );
     ~HdfFile();
@@ -129,10 +129,8 @@ class HdfGroup
 
     inline HdfGroup group( const std::string &groupName ) const;
     inline HdfDataset dataset( const std::string &dsName ) const;
-
     inline HdfAttribute attribute( const std::string &attr_name ) const;
     inline bool pathExists( const std::string &path ) const;
-
   protected:
     std::vector<std::string> objects( H5G_obj_t type ) const;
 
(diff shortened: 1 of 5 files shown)


mdal/frmts/mdal_hdf5.hpp
```diff
@@ -66,6 +66,7 @@ class HdfFile
       Create
     };
     typedef HdfH<H5I_FILE> Handle;
+    typedef std::shared_ptr<Handle> SharedHandle;
```
PeterPetrik, line 69: do we need new `typedef` personally for me it is easier to read the code when shared_ptrs are expanded everywhere.

mdal/frmts/mdal_hdf5.hpp
```diff
@@ -75,16 +76,21 @@ class HdfFile
...
     inline bool pathExists( const std::string &path ) const;
     std::string filePath() const;
 
   protected:
-    std::shared_ptr<Handle> d;
+    SharedHandle d;
     std::string mPath;
 };
 
+typedef std::shared_ptr<HdfFile> HdfFileShared;
```
PeterPetrik, line 92: do we need new `typedef` personally for me it is easier to read the code when shared_ptrs are expanded everywhere.

mdal/frmts/mdal_hdf5.hpp
```diff
@@ -132,8 +134,17 @@ class HdfGroup
   protected:
     std::vector<std::string> objects( H5G_obj_t type ) const;
 
+  private:
+    HdfGroup( HdfFile::SharedHandle file, const std::string &path );
+    HdfGroup( std::shared_ptr<Handle> handle, HdfFile::SharedHandle file );
+
+  private:
+    HdfFile::SharedHandle mFile; //must be the last destroyed
```
PeterPetrik, line 142: I think the commend it misleading, I would remove it. If you want to add comment, I would say do not destroy from this class

tests/test_xmdf.cpp
```diff
@@ -299,6 +299,27 @@ TEST( MeshXmdfTest, withReferenceTime )
...
+  std::string path = test_file( "/xmdf/withReferenceTime/hydraul_006.2dm" );
+  EXPECT_EQ( MDAL_MeshNames( path.c_str() ), "2DM:\"" + path + "\"" );
+  MDAL_MeshH m = MDAL_LoadMesh( path.c_str() );
+  ASSERT_NE( m, nullptr );
+  MDAL_M_LoadDatasets( m, tmpXmdf.c_str() );
+  ASSERT_EQ( 12, MDAL_M_datasetGroupCount( m ) );
+
+  MDAL_CloseMesh( m );
+
+  deleteFile( tmpXmdf );
```
PeterPetrik, line 318: This test properly tests it only on Windows, right? Also on Windows I fear it may fail, because windows may have some delay for garbage collection and freeing handles. I am wondering if it is not better approach to check the same? Like check with HDF library that all the handles are closed?

PeterPetrik (commented):
Looks in general good, left some comments to discuss

tests/test_xmdf.cpp
vcloarec: > This test properly tests it only on Windows, right?


yes


> windows may have some delay for garbage collection and freeing handles

Not sure to understand, do you mean that the HDF lib will have some delay to freeing the handle under Windows? Or windows will have so delay to really delete the file?


> Like check with HDF library that all the handles are closed?

Check from the test with HDF lib, is it possible without the file handle?

mdal/frmts/mdal_hdf5.hpp
vcloarec: yes, the comment should be "must be declared before `std::shared_ptr<Handle> d` to be sure it will be the last "

PeterPetrik (approved):

PR #426, opened 2022-09-07
fix lock file for cf drivers

vcloarec:
fix https://github.com/qgis/QGIS/issues/49215

### DIFF
--- a/mdal/frmts/mdal_cf.cpp
+++ b/mdal/frmts/mdal_cf.cpp
@@ -472,10 +472,12 @@ bool MDAL::DriverCF::canReadMesh( const std::string &uri )
   }
   catch ( MDAL_Status )
   {
+    mNcFile.reset();
     return false;
   }
   catch ( MDAL::Error )
   {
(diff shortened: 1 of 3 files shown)


PeterPetrik (approved):
not sure what is the change 🤷

rduivenvoorde:
@vcloarec should/can I test (my issue)? I'm actually not aware how mdal is based/build in QGIS? Do you pull source in QGIS sometimes? I see: https://github.com/qgis/QGIS/tree/master/external/mdal which does not seem to have this fix?

saberraz:
hi Richard, when you compile your qgis, you need to use this flag: `-DWITH_INTERNAL_MDAL=FALSE`

Then compile MDAL (https://github.com/lutraconsulting/MDAL#linux), back to QGIS, re-run `ninja` or `make`. If your MDAL is installed in a system path, it should be picked up by QGIS.

rduivenvoorde:
Ah, thanks!

Issue #428, opened 2022-09-14
Problem with UGRID to load file if directory contains special char


ccoulet:
Hi
I'm an old school informatic user so I often create directory and file without any spaces but with shared space, I'm not always the decider...
I recently encountered a problem with a netcdf ugrid file which is well open by QGIS/MDAL when located on my computer and where a crash occur if I open the same file located on a server.
The only difference I saw is the name of the path which contains spaces on the server architecture.
As other files (gpkg, shp, ...) are well opened by QGIS, I suspect this in an internal problem in MDAL but this need to be confirmed and potentially corrected
Thanks in advance

ccoulet:
Hi Vincent
Latest version of Qgis (3.26.2)
Load from Data Source Manager
For the path (with some adaptation for confidentiality reasons)
It works with this path
D:\Documents\coulet\Travail\xxxx_xxxxxxxxxxxxxxxxx\xx_xxxx\Calculs_xxx\Resultats\max_x_xxxx_x.xxxx.nc

and crash with this one:
D:\Documents\coulet\xxxx\xxx xxxxxxxxxx - 04-Données xxxxxxxxxxxxx xx xxxxxx\03_Travail\01_xxxxxxxxxx_xxxxxxxx_xxx\Calculs_xxx\Resultats\max_x_xxxx_x.xxxx.nc

ccoulet:
I realized that there is also a "é" which could probably be a potential source of problem.

vcloarec:
I confirm an issue under Windows, not leading to a crash, but the file is not recognized as a UGRID file when the path contains "é". But with space it is OK.

ccoulet:
Thanks for investigate this problem

vcloarec:
@ccoulet , #429 could fix your issue, but I am not complety sure...

ccoulet:
I will see in the next version.
Thanks for your investment.

vcloarec:
@ccoulet Have you try with new version of QGIS?

ccoulet:
Sounds working!
Thanks

PR #429, opened 2022-09-15
fix netCdf derived driver if there is special char on the file name (Windows)

vcloarec:
could fix https://github.com/lutraconsulting/MDAL/issues/428


mdal/mdal_utils.cpp
```diff
@@ -400,6 +401,40 @@ std::string MDAL::trim( const std::string &s, const std::string &delimiters )
...
+{
+  std::string ret = utf8FileName;
+#if defined(WIN32)
+  ret = utf8ToWin32Recode( utf8FileName );
+#endif
+  return ret;
+}
+
+#if defined(WIN32)
+std::string MDAL::utf8ToWin32Recode( const std::string &utf8String )
```
PeterPetrik, line 414: is this something else than https://github.com/lutraconsulting/MDAL/blob/fa15bfa4a4bc5c15195cead36ee181bcd93302e4/mdal/mdal_utils.cpp#L40 ? or different? Maybe we can make it common if it does the same thing?

mdal/mdal_utils.cpp
```diff
@@ -20,6 +20,7 @@
 #define UNICODE
 #include <locale>
 #include <codecvt>
+#include <stringapiset.h>
```
PeterPetrik, line 23: `#if defined(WIN32)`

mdal/mdal_utils.hpp
```diff
@@ -139,6 +139,12 @@ namespace MDAL
     const std::string &s,
     const std::string &delimiters = " \f\n\r\t\v" );
 
+  std::string systemFileName( const std::string &utf8FileName );
+
+//#if defined(WIN32)
```
PeterPetrik, line 144: `#if defined(WIN32)` or even better make it static function in mdal_utils.cpp and not add it to header at all.

mdal/mdal_utils.cpp
```diff
@@ -400,6 +401,40 @@ std::string MDAL::trim( const std::string &s, const std::string &delimiters )
   return ltrim( rtrim( s, delimiters ), delimiters );
 }
 
+std::string MDAL::systemFileName( const std::string &utf8FileName )
+{
+  std::string ret = utf8FileName;
```
PeterPetrik, line 406: I would put this to `#else`

PeterPetrik (approved):

PR #433, opened 2022-10-10
Bump googletest to fix build issue

caspervdw:
I got an error during buillding the included library "googletest" with GCC 11.

Bumping googletest fixed it.

Ref. ['dummy' may be used uninitialized](https://stackoverflow.com/questions/69935158/dummy-may-be-used-uninitialized) and https://github.com/google/googletest/pull/3024


PeterPetrik (approved):

PeterPetrik:
Thanks!

PR #453, opened 2023-04-13
[PLY] Fix for 450

### LINKED ISSUE #450: [PLY] Saved mesh with PLY format seems incompatible with some other mesh reader
see https://github.com/qgis/QGIS/issues/52467

### LINKED ISSUE #452: [PLY] Save in binary mode
Currently - MDAL can only save ply files in text mode. It would be good to also allow binary mode if possible.

Assign to me.

runette:
fixes #450 
fixes #452 

Updates to the PLY driver :

- Changes the vertices names to lower case : "x", "y", "z"
- Adds the ability to save in binary mode
- changes the default behaviour to save in binary mode to avoid precision issues
- adds the "format" metadata item to maintain round-trip consistency and allow the format to be changed
- fixes a bug in the handling of CRS data

The resulting binary and text files have been tested in CloudCompare and Blender and open succesfully


runette:
@vcloarec @PeterPetrik - please review

mdal/frmts/mdal_ply.cpp
```diff
@@ -577,15 +580,44 @@ void MDAL::DriverPly::save( const std::string &fileName, const std::string &mesh
     }
   }
 
+  libply::Metadata meta;
+
+  meta.emplace( "crs", mesh->crs() );
+  MDAL::Metadata metadata = mesh->metadata;
```
vcloarec, line 586:
```suggestion
  const MDAL::Metadata &metadata = mesh->metadata;
```

mdal/frmts/mdal_ply.cpp
```diff
@@ -577,15 +580,44 @@ void MDAL::DriverPly::save( const std::string &fileName, const std::string &mesh
...
+  libply::File::Format format = libply::File::Format::BINARY_LITTLE_ENDIAN;
+
+  const std::unordered_map<std::string, libply::File::Format> FORMAT_MAP =
+  {
+    { "ascii", libply::File::Format::ASCII },
+    { "binary_big_endian", libply::File::Format::BINARY_BIG_ENDIAN },
+    { "binary_little_endian", libply::File::Format::BINARY_LITTLE_ENDIAN }
+  };
+
+  for ( auto it = metadata.begin(); it != metadata.end(); ++it )
```
vcloarec, line 596:
```suggestion
  for ( auto it = metadata.cbegin(); it != metadata.cend(); ++it )
```

mdal/frmts/mdal_ply.cpp
```diff
@@ -577,15 +580,44 @@ void MDAL::DriverPly::save( const std::string &fileName, const std::string &mesh
...
+  {
+    std::pair< std::string, std::string > item = *it;
```
vcloarec, line 598:
```suggestion
    const std::pair< std::string, std::string > &item = *it;
```

mdal/frmts/mdal_ply.cpp
```diff
@@ -577,15 +580,44 @@ void MDAL::DriverPly::save( const std::string &fileName, const std::string &mesh
     }
   }
 
+  libply::Metadata meta;
+
+  meta.emplace( "crs", mesh->crs() );
+  MDAL::Metadata metadata = mesh->metadata;
+  libply::File::Format format = libply::File::Format::BINARY_LITTLE_ENDIAN;
+
+  const std::unordered_map<std::string, libply::File::Format> FORMAT_MAP =
```
vcloarec, line 589: any reasons to be upper case?

vcloarec (commented):
Nothing particular to say except some `const` . Do you think it is possible to avoid to  decrease the test coverage?

runette:
> Nothing particular to say except some `const` . Do you think it is possible to avoid to decrease the test coverage?

I will take a look at the detailed comments. On the general question - I was worried about that as well (even though it is only 0.02%) and did look at it - but the coverage calculation is a bit strange as it is complaining about 4 lines of code that are definitely tested!

I will see what I can do. I think it is because big-endian is not tested. I will try that.

mdal/frmts/mdal_ply.cpp
runette: Nope. That is a brain_fart! I think it was an unconscious association with the line above :(

vcloarec:
I am ok with the coverage decrease (only -0.02%), but I am concerned about that we are under 90%, even it is not you fault here.

runette:
> I am ok with the coverage decrease (only -0.02%), but I am concerned about that we are under 90%, even it is not you fault here.



> I am ok with the coverage decrease (only -0.02%), but I am concerned about that we are under 90%, even it is not you fault here.

I will have a look at it this evening and see if I can find a way to get it above 90

runette:
@vcloarec  @PeterPetrik 

It should now be green - I am not sure why it is not.

I did not get coverage to 90% but got it going the right way.

I increased test coverage to include big_endian files (and in the process discovered and fixed two bugs - so I guess testing is good after all :). I also deleted some unreachable code.

I tested the outputs from this driver in all formats with CloudCompare and Blender and it seems to be working

I propose that it is merged

vcloarec (approved):

PR #458, opened 2023-05-24
Prepare 1.0.3

vcloarec:
for 1.0.3 release


mdal/3rdparty/libplyxx/libplyxx.cpp
```diff
@@ -51,8 +51,8 @@ SOFTWARE.
 #include <string>
 #include <iostream>
 
-#define BIG_ENDIAN 1
-#define LITTLE_ENDIAN 0
+#define PLYXX_BIG_ENDIAN 1
```
vcloarec, line 54: @runette 
I am trying to push new version of MDAL to QGIS.
These macros are already used, so I purpose to rename it.
is it OK?

PeterPetrik (approved):

mdal/3rdparty/libplyxx/libplyxx.cpp
runette: No problem

PR #467, opened 2023-09-25
Add support for 3d extent

ptitjano:
This introduces a new function `MDAL_M_extent3D` to get the 3D extent. This also adds support for all the relevant drivers.

The different unit tests are updated.


vcloarec:
Hi, I think it is a good move to access to the vertical extent of the mesh frame. But I also think it could be confusing to use 3D extent. Indeed, mesh are not explicitly 3D  entities. Some are (not  a lot), some are only 2.5D , and other are only 2D.
Also, some dataset could have some elevation values, that could be included in the vertical extent of the mesh.

Here, you consider that the Z value of the vertices is the third dimension, but it is not always the case. 
Some meshes have their verticale positions expressed by elevation on face, and values are in the dataset. 

So, I think the term 3D extent is not the good one in your case, and you should not have such function.

If your need is only for the Z values of the vertices, I think you should have an API function `MDAL_M_vertices_vertical_extent`. Then following your need you can use both this function and `MDAL_M_extent`.

runette:
I am not sure that I totally agree with @vcloarec about some of the comments. But I also have issues with the PR as well.

1. As someone who uses MDAL almost exclusively in an actual 3D environment, getting the 3D extent in one call would be very useful. I think that the intent of the PR is good and the suggestion of a 2D extent and a vertical extent is both less good and also does not introduce any simplification.
2.  I have a deep conceptual problem with saying that GIS objects are NOT intrinsically 3D (and indeed the whole idea of 2.5D GIS). Yes, I know that these concepts are canonical in GIS but I also think such ideas are fundamentally wrong and am in the process of writing a paper about this. Yes, many meshes are manifolds of order 2( i.e. 2D manifolds) (and indeed some are 1D manifolds) but by dint of being GIS objects those manifolds are always embedded into real 3D space. I could go on about this at ridiculous length ( :) ) but the takeaway is that "3D Extent" or 3D bounding box always has meaning for GIS mesh datasets, as indeed is indicated by MDAL's default dataset of "bed elevation", even if in some cases the answer is trivially different from the "2D extent".
3. That said, there are problems with the PR particularly as above about the assumptions of what is the z dimension and also what. is meant by "the relevant drivers are updated"?

I would suggest that @ptitjano should provide more information about the intended operation of the method, particularly in cases where there is no Z data. My feeling is that where there is a Z dataset this should be used (and my understanding is that in this case the Z dataset should be correlated with Bed Elevation). Where there is no Z dataset but there is Bed Elevation data then this should be used. Otherwise, the z dimension in the 3D Extent should be zero or NaN (I would prefer the former). But the important thing is that this is all documented before review and merge.

ptitjano:
Thanks @vcloarec and @runette for the comments.

I have been trying to visualize mesh data in QGIS 3D and improve its support. More generally, I have been trying to add a `extent3D()` method to all QGIS providers. This new api would then be consumed by QGIS 3D to populate the configuration of the initial view and improve data loading (at the moment QGIS 3D does not any idea of the vertical extent of the data and it needs to make some weird assumptions to properly load them.)
As far as meshes are concerned, this means that I need to have a MDAL function to retrieve the 3D extent (bounding box) of a mesh. It would have the following behavior: 

- 2D: `{xMin, yMin, NaN, xMax, yMax, NaN}`. 
- 3D: `{xMin, yMin, zMin, xMax, yMax, zMax}`;

I do not have a deep knowledge of MDAL or Meshes. I would like to have the 3D bounding box of a mesh, whether the 3D information is stored in the vertices, the faces or as values in the dataset. Does it make sense ? If that's the case, how could it be achieved ? If not, what other options could be tried?

PeterPetrik:
@ptitjano  check this https://github.com/qgis/QGIS-Enhancement-Proposals/issues/158

vcloarec:
@runette , I don't want to say that GIS object is not 3D, but either that is is difficult to express 3D extent of a MDAL mesh through only one methods. Indeed, there are many ways to define the 3D of a mesh, that can depend on: 
- the Z value of vertices if defined ,
- if elevation of elements are defined in a dataset group
- what is 3D extent, only the mesh frame or including some dataset as water surface for hydrodynamics models,
- what the user of MDAL want to consider to define the 3D extent, including some dataset group and which one.
- if the mesh is a 3D stack or not (thanks @PeterPetrik  to remember this type of mesh)

Indeed, ideally, this 3D extent function in the MDAL API should express directly the 3D extent of the mesh for all mesh regardelss of the structure of this mesh and its dataset. But I don't think it is possible considering all the format supported. PLY format is one of the format for which it is obvious and could be easy to have this methods because the 3D extent will depends only on the vertices, but for some other formats it is not obvious.

That's why I purpose to use some method that return more explicitly what we need, and the user of MDAL will use what he wants to define the 3D extent following its needs. The responsibility of defining what is the 3D extent is not MDAL but the user.

@ptitjano , If the aim of this PR is to have the 3D extent of what QGIS 3D will render, it will depends on what you want to render : a terrain or a dataset. If it is the terrain, the vertical extent of the vertices  + the extent of the mesh should make the job (be aware that QGIS 3D render on map coordinate, not with provider coordinate). If you want to render dataset, the verticale extent will depends of this dataset. Moreover, I think that must be handle in the QGIS side, not in the MDAL side.

I hope that could help.

runette:
Indeed.

The disconnect here is not that I do NOT understand the problem - in fact it is because I actually fully understand the problem since, when representing mesh data in a native 3D space (VR), I have to address the issues every time.

TLDR;
1. I do not think that creating an explicit 3D extent is too complicated - as long as the "the user has created a non-standard Z dataset" corner case can be excluded. It is already for the X and Y dimensions.
2. I think that doing so would be extremely valuable.

My main starting point is that a data abstraction layer is there not to avoid the hard problems but to address the hard problems so that the consumer does not have to!

In particular, in this case, the issues that you address are the critical ones but they can be addressed and by addressing them MDAL would be providing a valuable Data Abstraction Layer service:

1. All GIS meshes are 1D, 2D or 3D manifolds embedded in 3D space,
2. If 1D or 2D manifolds then the vertices are 3D points (see 1.) with the Z dimension either
- Implicit ( the meaning of which depends effectively on the CRS but can be represented usually as 0)
- explicit and defined as Z coordinates
- explicit and defined as Bed Elevation and not Z coordinates (?)
- explicit and defined as some other dataset. NOTE - it would be perfectly acceptable (and to my mind preferable) to say that this is not supported.
3. 3D manifolds - called 3D Layered Mesh in the MDAL documentation and Stacked Data above and Volumetric mesh data in most of the literature (and, curiously, in the MDAL C API)
- Have a explicit 3D extent since they are made up of voxels explicitly located in 3D space
- But, you are about to say, the Z dimension is defined on the Dataset and not the mesh!
- But, this is just to trivially say that the actual 3D extent of a mesh with Layered mesh Data (or volumetric Datasets) is the AABB of the individual AABBs of the individual datasets

PeterPetrik:
Hi, thank you all for valuable discussion. I created issue to not forget about this, and closing this PR in the meantime to clean PR queue.

PR #470, opened 2023-11-22
Mike21 Driver

JanCaha:
Driver for Mike21 format:  https://www.xmswiki.com/wiki/SMS:MIKE_21_*.mesh.

Currently there is some code duplication with 2DM format, which could be eliminated by making functions `_parse_vertex_id_gaps()` and `_persist_native_index()` part of MDAL API. Also the class that extends `MemoryMesh` to store vertexIDs that might not be continuous could be generalized and shared by Mike21 and 2DM format.


PeterPetrik (commented):
looks good

PeterPetrik:
> `_parse_vertex_id_gaps()` and `_persist_native_index()` part of MDAL API

definitely not to MDAL API, if you want to can add it to mdal_utils , but it is not necessary

JanCaha:
Fix the issues, you mentioned @PeterPetrik . Also realized that there are few more validations that can be done on the file, so I added them + tests.

PeterPetrik (approved):

JanCaha:
There was an issue where the new driver was trying to load another format, due to too simple `canReadMesh()`.  Now better checks are performed to ensure that the Mike21 file can be read prior to reading. Which avoids the issue.

mdal/frmts/mdal_mike21.cpp
```diff
@@ -128,6 +128,39 @@ bool MDAL::DriverMike21::canReadMesh( const std::string &uri )
...
     return false;
   }
+
+  in.clear();
+  in.seekg( 0, std::ios::beg );
+
+  size_t vertexCount = getVertexCount( line );
+  size_t faceCount = 0;
+
+  size_t lineNumber = 0;
```
PeterPetrik, line 138: I do not like this approach , if the file is massive you pretty much read/parse it fully only to say it is MIKE 21 file... 

canReadMesh is called many times for each file (even read by other drivers) so it should be fast

PeterPetrik:
looks good; @JanCaha you need to add new driver to 
- https://github.com/lutraconsulting/MDAL/tree/master/docs/source/drivers
- https://github.com/lutraconsulting/MDAL/blob/master/README.md

mdal/frmts/mdal_mike21.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include "mdal_memory_data_model.hpp"
+#include "mdal.h"
+#include "mdal_driver.hpp"
+
+namespace MDAL
+{
+  class MeshMike21: public MemoryMesh
+  {
+    public:
+      MeshMike21( size_t faceVerticesMaximumCount,
```
vcloarec, line 23: is it necessary to use this new derived class ? I don't see in the code where the methods of this class are used.

mdal/frmts/mdal_mike21.hpp
```diff
@@ -0,0 +1,104 @@
...
+      bool canReadMesh( const std::string &uri ) override;
+      std::unique_ptr< Mesh > load( const std::string &meshFile, const std::string &meshName = "" ) override;
+      void save( const std::string &fileName, const std::string &, Mesh *mesh ) override;
+
+      std::string saveMeshOnFileSuffix() const override;
+
+    private:
+      std::string mMeshFile;
+      // regex for header line in form of - integer string
+      std::regex mRegexHeader2011 = std::regex( "(\\d+)\\s+(.+)(\\s+)?" );
```
vcloarec, line 92:
```suggestion
     const std::regex mRegexHeader2011 = std::regex( "(\\d+)\\s+(.+)(\\s+)?" );
```

mdal/frmts/mdal_mike21.hpp
```diff
@@ -0,0 +1,104 @@
...
+      // regex for header line in form of - integer integer integer string
+      std::regex mRegexHeader2012 = std::regex( "(\\d+)\\s+(\\d+)\\s+(\\d+)\\s+(.+)(\\s+)?" );
```
vcloarec, line 94:
```suggestion
      const std::regex mRegexHeader2012 = std::regex( "(\\d+)\\s+(\\d+)\\s+(\\d+)\\s+(.+)(\\s+)?" );
```

mdal/frmts/mdal_mike21.hpp
```diff
@@ -0,0 +1,104 @@
...
+      // regex for element header line - integer integer integer(code with two number)
+      std::regex mRegexElementHeader = std::regex( "(\\d+)\\s+(\\d)\\s+(\\d{2})(\\s+)?" );
```
vcloarec, line 96:
```suggestion
      const std::regex mRegexElementHeader = std::regex( "(\\d+)\\s+(\\d)\\s+(\\d{2})(\\s+)?" );
```

mdal/frmts/mdal_mike21.hpp
JanCaha: This just copies the approach used in 2DM driver and it should allow the use of the mesh structure later. No direct use for those fuctions right now. But for 2DM driver it is rather similar, those functions are only used in `ASCII_DAT` format.

PeterPetrik: I think it is fine... easier to understand  it is in-Memory?

JanCaha: Yes

PeterPetrik (approved):

Issue #479, opened 2024-03-16
Selafin header record not generated correctly


nstrahl:
The default header record created in MDAL::DriverSelafin::save and written has a bug. The wrong std::string constructor overload is being used, so the header is populated with garbage (undefined) bytes. This crashes the telemac2d.py script when running a telemac simulation.

PeterPetrik:
fixed in https://github.com/lutraconsulting/MDAL/pull/480

PR #480, opened 2024-03-16
Fix: Bad bytes in header record

nstrahl:
Fixes issue #479

The correct std::string constructor is now called.


PeterPetrik:
Hi @nstrahl , thanks for fixing the bug!

PeterPetrik (approved):

PR #484, opened 2024-03-21
Fix selafin special characters

### LINKED ISSUE #483: Writing a selafin mesh dataset to a file path with special characters fails under windows
Using file paths for non-ANSI characters (äöü) is not handled properly by the selafin driver in MDAL::SelafinFile::addDatasetGroup.
This can be quickly verified in QGIS, where the native:tinmeshcreation algorithm will run without error. However, the resulting mesh will only contain the mesh structure without the "elevation" dataset.

nstrahl:
Fixes #483

First off, a regression test (WriteDatasetSpecialCharacters) was written so that the issue at hand could be reproduced.
Then the issue was fixed. Some instances of std::string had to be converted to std::wstring followed by windows-specific system calls.
The solution is analogous to the way cross-platform strings are currently handled in mdal_utils.h/.cpp.

### DIFF
--- a/mdal/mdal_utils.cpp
+++ b/mdal/mdal_utils.cpp
@@ -111,7 +111,7 @@ bool MDAL::deleteFile( const std::string &path )
     std::wstring wStr = converter.from_bytes( path );
     return DeleteFileW( wStr.c_str() ) != 0;
 #else
-    return remove( path.c_str() ) == 0;
+    return std::remove( path.c_str() ) == 0;
 #endif
   }
 
--- a/mdal/mdal_utils.cpp
+++ b/mdal/mdal_utils.cpp
@@ -102,31 +102,31 @@ std::string MDAL::readFileToString( const std::string &filename )
   return "";
 }
 
-bool MDAL::deleteFile(const std::string& path)
+bool MDAL::deleteFile( const std::string &path )
 {
-    if (MDAL::fileExists(path))
-    {
-#ifdef _MSC_VER
-      std::wstring_convert< std::codecvt_utf8_utf16< wchar_t > > converter;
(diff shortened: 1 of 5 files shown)


mdal/mdal_utils.hpp
```diff
@@ -71,6 +71,12 @@ namespace MDAL
   std::string pathJoin( const std::string &path1, const std::string &path2 );
   std::string readFileToString( const std::string &filename );
 
+  //! Deletes a file. Returns true on success, false otherwise
+  bool deleteFile( const std::string &path );
+
+  //!renames a file. Returns true on success, false otherwise
```
PeterPetrik, line 77: //!renames -> //! Renames

mdal/mdal_utils.cpp
```diff
@@ -102,6 +102,34 @@ std::string MDAL::readFileToString( const std::string &filename )
...
+bool MDAL::deleteFile( const std::string &path )
+{
+  if ( MDAL::fileExists( path ) )
+  {
+#ifdef _MSC_VER
+    std::wstring_convert< std::codecvt_utf8_utf16< wchar_t > > converter;
+    std::wstring wStr = converter.from_bytes( path );
+    return DeleteFileW( wStr.c_str() ) != 0;
+#else
+    return remove( path.c_str() ) == 0;
```
PeterPetrik, line 114: std::remove

PeterPetrik (commented):
Thanks for fixing the stuff!

PeterPetrik (approved):

nstrahl:
Thanks! Could you make a patch version (1.1.1) for the last fixes? Or would that be too soon according to the roadmap?

PeterPetrik:
@uclaros ?

PR #497, opened 2025-02-05
Fix 2dm precision

JanCaha:
Specify precision for 2DM coordinate values based on magnitude of the coordinate.


.github/workflows/coverage_tests.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
```
uclaros, line 6: Maybe switch from ubuntu-latest to 22.04?

uclaros (approved):

PR #500, opened 2025-04-14
Fix out of bounds access when reading dat reference time

nyalldawson:
We can't store an 8 byte result into a 4 byte float value


uclaros (approved):
Sorry for the delay, I was not watching the repo :roll_eyes: 

I guess it would now make sense to move the `float time;` declaration into the `CT_TS` case, too?

nyalldawson:
@uclaros 
> I guess it would now make sense to move the float time; declaration into the CT_TS case, too?

Done, thanks!

PR #502, opened 2025-05-08
Cmake cleanups

nyalldawson:


uclaros (approved):
find sqlite seems to work again now

PR #507, opened 2025-05-08
Fix leak of Library shared data

nyalldawson:
The shared data pointer was never deleted when the ref count hit 0

### DIFF
--- a/mdal/mdal_utils.cpp
+++ b/mdal/mdal_utils.cpp
@@ -1112,13 +1112,18 @@ MDAL::Library::Library( std::string libraryFile )
 MDAL::Library::~Library()
 {
   d->mRef--;
+  if ( d->mRef == 0 )
+  {
+    if ( d->mLibrary )
+    {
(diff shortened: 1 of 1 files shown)


mdal/mdal_utils.cpp
```diff
@@ -1112,13 +1112,17 @@ MDAL::Library::Library( std::string libraryFile )
 MDAL::Library::~Library()
 {
   d->mRef--;
+  if ( d->mRef == 0 )
```
uclaros, line 1115: I think it would be slightly more readable like:
```cpp
  if ( d->mRef == 0 )
  {
    if ( d->mLibrary )
    {
#ifdef _WIN32
      FreeLibrary( d->mLibrary );
#else
      dlclose( d->mLibrary );
#endif
    }
    delete d;
  }
```

nyalldawson: Ok, done

uclaros (approved):

PR #514, opened 2025-10-08
Fix HDF5 dataset loading (handle error codes)

### LINKED ISSUE #63216: HDF5 cannot be opened through drag-drop
### What is the bug or the crash?

I have a HDF5 which I can open through Data Source Manager > Raster. But when I try to drag and drop the file in QGIS, it comes up with the following error:

Image

CC @timlinux 

### Steps to reproduce the issue

- Find a HDF5
- Use Data Source Manager > Raster and it will add the file content correctly to QGIS
- Try to drag and drop the file, and you will see the above error message.

### Versions

3.44 and master

### Supported QGIS version

- [x] I'm running a supported QGIS version according to [the roadmap](https://www.qgis.org/en/site/getinvolved/development/roadmap.html#release-schedule).

### New profile

- [x] I tried with a new [QGIS profile](https://docs.qgis.org/latest/en/docs/user_manual/introduction/qgis_configuration.html#working-with-user-profiles)

### Additional context

_No response_

ViperMiniQ:
While opening the HDF5 dataset, we check for error codes and possible negative values, so we do not end up static casting them to unsigned ints.

Fixes qgis/QGIS#63216

### DIFF
--- a/mdal/frmts/mdal_hdf5.cpp
+++ b/mdal/frmts/mdal_hdf5.cpp
@@ -193,7 +193,17 @@ hid_t HdfDataset::id() const { return d->id; }
 std::vector<hsize_t> HdfDataset::dims() const
 {
   hid_t sid = H5Dget_space( d->id );
-  std::vector<hsize_t> ret( static_cast<size_t>( H5Sget_simple_extent_ndims( sid ) ) );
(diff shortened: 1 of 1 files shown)


uclaros (approved):
Thanks!

PR #517, opened 2025-10-20
[ply] Check the results of dynamic casts

nyalldawson:
Gracefully log errors and don't crash when expected casts fail. Makes driver more tolerant to malformed data files and avoids coverity scan errors.

### DIFF
--- a/mdal/3rdparty/libplyxx/libplyxx.cpp
+++ b/mdal/3rdparty/libplyxx/libplyxx.cpp
@@ -355,6 +355,11 @@ namespace libply
         const size_t listLength = std::stoi( m_tokens[t_idx] );
         t_idx++;
         ListProperty *lp = dynamic_cast<ListProperty *>( &elementBuffer[e_idx] );
+        if ( !lp )
+        {
+          MDAL_SetStatus( MDAL_LogLevel::Error, MDAL_Status::Err_InvalidData, "Element is not a List Property" );
+          return;
+        }
         lp->define( p.type, listLength );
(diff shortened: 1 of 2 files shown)


mdal/3rdparty/libplyxx/libplyxx.cpp
```diff
@@ -559,7 +565,7 @@ namespace libply
       else
       {
         auto &convert = p.writeConvertFunction;
-        ListProperty *lp = dynamic_cast<ListProperty *>( &buffer[e_idx] );
+        ListProperty *lp = static_cast<ListProperty *>( &buffer[e_idx] );
```
uclaros, line 568: Why `static_cast`? Can't this be a `ScalarProperty` here too?

mdal/3rdparty/libplyxx/libplyxx.cpp
```diff
@@ -604,7 +610,7 @@ namespace libply
       else
       {
         auto &cast = p.writeCastFunction;
-        ListProperty *lp = dynamic_cast<ListProperty *>( &buffer[e_idx] );
+        ListProperty *lp = static_cast<ListProperty *>( &buffer[e_idx] );
```
uclaros, line 613: Same as above?

mdal/3rdparty/libplyxx/libplyxx.cpp
nyalldawson: There's an "is list" check above, I assumed that was sufficient

uclaros (approved):

PR #523, opened 2026-01-27
Change GDALmetadata type from char** to CSLConstList

mdsumner:
CSLConstList for GDALMetadata - I just keep tripping on this because of GDAL 3.13+

hth!


landryb:
fwiw, just stumbled on this very same issue after upgrading gdal to 3.13, and i can confirm it fixes the build breakage. also see #522 and #527

uclaros:
Thanks, #522 got merged since it was the downstream QGIS change

PR #529, opened 2026-05-12
Fix doc build

nicogodet:


nicogodet:
It seems it can not be fixed without an update of osgeo docker image or sacrificing functionalities

nicogodet:
Closing in favor of #530

PR #525, opened 2026-03-27
[sww] Read EPSG, starttime, timezone and centroid (_c) datasets

stoiver:
- Read global `epsg` attribute and set mesh CRS via setSourceCrsFromEPSG()
- Read global `starttime` (Unix epoch seconds) as reference time on all dataset groups, and `timezone` as mesh metadata
- Support centroid-based variables (suffix `_c`, dimension number_of_volumes): scalar and vector, static and time-dependent, stored as DataOnFaces
- Generalise readScalarGroup/readVectorGroup to accept nValues and MDAL_DataLocation so vertex and face groups share one code path
- Fix time units in readVectorGroup to use RelativeTimestamp::seconds, consistent with readScalarGroup
- Add test file merimbula_basic_mesh.sww (EPSG:32755, has _c variables)
- Update Merimbula and Catchment test cases accordingly

PR #534, opened 2026-07-09
Fix mac ci

uclaros:
Fixes the macos workflow.
`macos-13` runners are retired, so replaced the existing job to use `macos-15-intel` instead, which shall live until late 2027.
A new `osx_tests_arm64` job was added using `macos-latest`.

This surfaced an issue with a test due to cpu arch change which is also fixed.


stoiver:
@uclaros it would be great to get  this PR was accepted so that I could use the update to progress my PR #525.

wonder-sk (approved):

